PR #4, opened 2020-11-20
Add ShellCheck GitHub Action

ceyhun:
Added https://github.com/ludeeus/action-shellcheck to lint shell scripts.
Fixed errors/warnings that it found.


cameronvoell (approved):
Changes look good, thanks for adding the shell check action

PR #6, opened 2020-11-25
Add release types

ceyhun:
Added notes about different types of releases, when they are made, what branches are involved and what are the differences between a regular one and betafix,hotfix in terms of checklist and automation script.

### DIFF
--- a/Releasing.md
+++ b/Releasing.md
@@ -160,3 +160,83 @@ cut a new release.
 <p>o Celebrate!</p>
 <!-- /wp:paragraph -->
 ```
+
+# Different types of releases
+
+## 1. Regular
+
+### When
+
+On Mondays, one week before main apps (WPiOS & WPAndroid) have cut their releases, every 2 weeks.
+
+### Branches
+
+For example when releasing gutenberg-mobile `1.11.0`.
+
+| Repo             | Cut From | Branch Name                        | Merging To      |
+| ---------------- | -------- | ---------------------------------- | --------------- |
+| gutenberg        | master   | rnmobile/release_1.11.0            | master          |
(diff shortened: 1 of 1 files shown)


Releasing.md
```diff
@@ -160,3 +160,83 @@ cut a new release.
...
+A fix is targeting a main app version that is not yet released (meaning the release branch is cut but it's still in beta) and a new gutenberg-mobile release is needed.
+
+### Branches
+
+For example when releasing gutenberg-mobile `1.11.1` while main apps version `22.2.0` is in beta which currently has gutenberg-mobile `1.11.0` in it.
+At the same time there could also be a regular release going on for example for gutenberg-mobile version `1.12.0`.
+
+| Repo             | Cut From                | Branch Name                        | Merging To                                                       |
+| ---------------- | ----------------------- | ---------------------------------- | ---------------------------------------------------------------- |
+| gutenberg        | rnmobile/release_1.11.0 | rnmobile/release_1.11.1            | master & (maybe also) rnmobile/release_1.12.0                    |
```
guarani, line 196: I understand that the "maybe also" here means that if the branch exists, merge back to that branch, is that correct? Does that apply to all the instances where you've added "maybe also"?

ceyhun: That's right, at the same time another release could be going on (or not) and it applies to all other instances.

guarani: Sounds good 👍 Thanks @ceyhun

cameronvoell (approved):
Thanks Ceyhun, updates here look good. I took a close look and all the info in here seemed correct, useful, and well laid out 👍

PR #12, opened 2020-12-10
Update branch name from main to trunk

guarani:
Since we're using the branch name `trunk` instead of `main`, this updates the docs to reflect that.


jd-alexander (approved):
LGTM 🚢

PR #17, opened 2021-01-05
Specify Xcode 12.1 is required on the command line

guarani:
On Xcode 12.2 I get this error:

```
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.
```

So specify that Xcode 12.1 is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ brew install jq
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


dcalhoun:
@Guarani I could be wrong, but I do not believe Xcode 12.1 is required any longer. I successfully ran the release script this past week with Xcode 12.4.

guarani:
Good catch @dcalhoun! I think this has gone stale since I also used Xcode 12.4 on the last release I did.
Closing the issue.

PR #20, opened 2021-02-02
Add --tab option to jq updating package.json files

ceyhun:
Gutenberg `package.json` files use tabs and `jq` converts them to spaces when modifying. Adding `--tab` option should fix this.

To test: 

In `gutenberg` or `gutenberg-mobile` run:

```sh
"jq" ".version = 9.99" "package.json" > "temp.json" --tab
"mv" "temp.json" "package.json"
```

Check if `package.json` has tabs.


fluiddot (approved):
LGTM 💯 !

I tested both running the command manually and running the script (only executing this part) and works fine.

I spotted that some files still have the 2 space indentation so in the next release they will be updated:
- gutenberg/packages/react-native-bridge/package.json
- gutenberg-mobile/package.json
- gutenberg-mobile/package-lock.json

PR #23, opened 2021-03-01
Update gutenberg to trunk from master

enejb:
Since the gutenberg repo switched from master to trunk for its main github branch. 
This PR makes the switch as well.


geriux (approved):
LGTM! Thanks for updating it!

PR #24, opened 2021-03-01
Gutenberg update the mobile label

enejb:
This PR updates the Gutenberg PR Label. 

The label got updated in https://github.com/WordPress/gutenberg/pull/29344/files

This PR makes it possible to run the script `./release_automation.sh` script without running into issues.

### DIFF
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -11,7 +11,7 @@
 #    d) WordPress-iOS: https://github.com/wordpress-mobile/WordPress-iOS
 # 2. Insure that each of your forked repos contains the PR labels specified below:
 GUTENBERG_MOBILE_PR_LABEL="release-process"
-GUTENBERG_PR_LABEL="Mobile App Android/iOS"
+GUTENBERG_PR_LABEL="Mobile App - i.e. Android or iOS"
 WPANDROID_PR_LABEL="gutenberg-mobile"
 WPIOS_PR_LABEL="Gutenberg integration"
 # 3. Ensure that each of your repos contains the target branch listed below:


cameronvoell (approved):
Nice catch @enejb changes look good. 

I wonder what's the best way to prevent this from breaking if people update repo Labels in the future. Let me know if you have any ideas, and thanks for the update!

enejb:
> I wonder what's the best way to prevent this from breaking if people update repo Labels in the future. Let me know if you have any ideas, and thanks for the update!

We could add a comment on the gutenberg repo or we could add a preflight check to the script. 
This way it would fail a bit earlier.

hypest:
I updated the GB repo label description to `Native mobile impl of the block editor. (Note: used in scripts, ping mobile folks to change)` as a first line of defence. Note: I was the one that recently changed the label, and wasn't even aware it's used in so many places/automation (release scripts, Slack channel automation).

mchowning:
> I wonder what's the best way to prevent this from breaking if people update repo Labels in the future. Let me know if you have any ideas, and thanks for the update!

If we had a number of different things to check it could be work creating a daily or even hourly CI job that goes through and checks things like the names of the relevant labels along with other things and pings us in Slack if anything has changed. Feels like overkill for just labels, but if there are more things, could be handy.

enejb:
I agree about the automation part. Lets not complicate things. :) 

I briefly looked into the github cli to see if we can do something about the label and failing early and I wasn't able to find a quick way to do this. 

I think we might be able to use the Github api to check for the existence (via https://api.github.com/repos/WordPress/gutenberg/labels) of labels.

PR #29, opened 2021-03-17
Add checklist item for updateing release incident spreadsheet

mchowning:
Seems like this will get forgotten if we don't add it to the checklist. wdyt?


chipsnyder (approved):
Yeah definitely good to add

mchowning:
Thanks @chipsnyder ! I have also updated [the 1.49.0 checklist](https://appreleasesp2.wordpress.com/2021/03/15/gutenberg-mobile-1-49-0-release-scenario/) to include it. If you have any concerns about that, just let me know.

chipsnyder:
> If you have any concerns about that, just let me know. 

No concerns but if you haven't checked already I updated the list when I approved. I probably should have mentioned it 😅

PR #30, opened 2021-03-17
Add a step related to update the Gutenberg reference

fluiddot:
This PR adds a new step in the release process in relation to update the Gutenberg reference after merging the Gutenberg PR into `trunk`.

The idea is that when we bring the release changes back to the development branches we assure that the Gutenberg reference is also updated and therefore that include the changes from Gutenberg.

### DIFF
--- a/Releasing.md
+++ b/Releasing.md
@@ -156,7 +156,15 @@ For the body of the post, just copy this checklist and again replace all occurre
 <!-- /wp:paragraph -->
 
 <!-- wp:paragraph -->
-<p>o Merge the Gutenberg PR to <code>trunk</code> and Gutenberg Mobile PR to <code>develop</code>.</p>
+<p>o Merge the Gutenberg PR to <code>trunk</code>.</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:paragraph -->
+<p>o Update the Gutenberg reference in the Gutenberg Mobile PR so it points to the Gutengerg PR merge commit in <code>trunk</code>.</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:paragraph -->
+<p>o Merge the Gutenberg PR to <code>trunk</code>.</p>
 <!-- /wp:paragraph -->
 
 <!-- wp:heading {"level":3} -->
--- a/Releasing.md
+++ b/Releasing.md
@@ -160,7 +160,7 @@ For the body of the post, just copy this checklist and again replace all occurre
 <!-- /wp:paragraph -->
 
 <!-- wp:paragraph -->
-<p>o Update the Gutenberg reference in the Gutenberg Mobile PR so it points to the Gutengerg PR merge commit in <code>trunk</code>.</p>
+<p>o Update the Gutenberg reference in the Gutenberg Mobile PR so it points to the Gutenberg PR merge commit in <code>trunk</code>.</p>
 <!-- /wp:paragraph -->
 
 <!-- wp:paragraph -->


Releasing.md
```diff
@@ -156,7 +156,15 @@ For the body of the post, just copy this checklist and again replace all occurre
 <!-- /wp:paragraph -->
 
 <!-- wp:paragraph -->
-<p>o Merge the Gutenberg PR to <code>trunk</code> and Gutenberg Mobile PR to <code>develop</code>.</p>
+<p>o Merge the Gutenberg PR to <code>trunk</code>.</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:paragraph -->
+<p>o Update the Gutenberg reference in the Gutenberg Mobile PR so it points to the Gutengerg PR merge commit in <code>trunk</code>.</p>
```
guarani:
```suggestion
<p>o Update the Gutenberg reference in the Gutenberg Mobile PR so it points to the Gutenberg PR merge commit in <code>trunk</code>.</p>
```

guarani (commented):
There's something I don't quite understand yet, so I wanted to clarify first before we make this change. 

One of the Last Day steps of the release is to merge the Gutenberg Mobile release PR to `trunk`. This means that on Gutenberg Mobile `trunk`, the submodule ref now points to the latest commit on the still un-merged Gutenberg release PR branch.

Then we do the "Bringing release changes back to development branches" steps and merge the contents of Gutenberg Mobile `trunk` into `develop`. Shouldn't this be enough to ensure `develop` now points to the released Gutenberg commit?

If we go with the changes proposed here to the checklist, the difference I see is that Gutenberg Mobile `develop` will now point to the merge commit instead of the latest commit on the Gutenberg release PR branch. Is that needed?

--- 

Looking at 1.48.1, when the PR was made to merge Gutenberg Mobile `trunk` into `develop`, was the submodule ref originally pointing to that latest commit on the Gutenberg release PR branch and it was then manually reverted to the same ref that develop had? Or by default was it already pointing to the same ref `develop` had? I read [this comment](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3253#issuecomment-797649541) but wasn't sure which was the case (or if it was something else).

Releasing.md
fluiddot: Thanks for the typo correction!

fluiddot:
>Then we do the "Bringing release changes back to development branches" steps and merge the contents of Gutenberg Mobile trunk into develop. Shouldn't this be enough to ensure develop now points to the released Gutenberg commit?

Sure, that would be enough but AFAIK the Gutenberg ref in `develop` should always point to a commit in `trunk`, not to a specific branch like the release one.

There's a reference to this topic in an internal post (p9ugOq-1bP-p2) just in case you want to check it, it's described in one of the last bullet points. Although it's true that it doesn't specifically say to have the reference pointing to the merge commit but I think that's the idea.

Besides usually there are conflicts with the reference when merging into `develop`, which can be tricky to solve if the reference in `develop` is ahead of the release one. In fact in this case, the only way to proceed would be to update the reference to the merge commit one because it will be more recent.

>If we go with the changes proposed here to the checklist, the difference I see is that Gutenberg Mobile develop will now point to the merge commit instead of the latest commit on the Gutenberg release PR branch. Is that needed?

Yeah that would be the difference, I think this would be needed if we want to keep that the Gutenberg reference always points to `trunk` as I commented above.

fluiddot:
>Looking at 1.48.1, when the PR was made to merge Gutenberg Mobile trunk into develop, was the submodule ref originally pointing to that latest commit on the Gutenberg release PR branch and it was then manually reverted to the same ref that develop had? Or by default was it already pointing to the same ref develop had?

Yeah it was originally pointing to the latest commit on the Gutenberg release PR, it was reverted to the `develop` one for solving the conflict.

In that case, what I should have done is merge the Gutenberg PR into `trunk` and then update the ref. That way we would have brought the changes that were done in `1.47.2` (the ones that weren't merged into `trunk`) into `develop`.

guarani (approved):
I see, I agree that ensuring Gutenberg Mobile's `develop` branch points to a Gutenberg submodule ref that is on Gutenberg's `trunk` is a good idea and a good strategy. When there are submodule ref conflicts when merging Gutenberg Mobile's `trunk` back to `develop`, it definitely makes sense that using the merge commit is better than using the latest commit on the Gutenberg release PR branch (since that would probably be older than what's on `develop`).
Thanks for clarifying! I think this looks good.

mchowning (approved):

dcalhoun (approved):

PR #39, opened 2021-04-14
Update the release automation script to update the version in build.gradle

enejb:
This PR updates the release_automation.sh script to account for the new android app release process as described in pbMoDN-2gI-p2. 

It does this by replacing the submodule (there is no submodule any more) ref updating with updating the string in `build.gradle`. 

# How has this been tested?
I created a helper script. 
```
#!/bin/bash

set -e

# Utils adapted from https://github.com/Homebrew/install/blob/master/install.sh
# string formatters
if [[ -t 1 ]]; then
    tty_escape() { printf "\033[%sm" "$1"; }
else
    tty_escape() { :; }
fi
tty_mkbold() { tty_escape "1;$1"; }
tty_underline="$(tty_escape "4;39")"
tty_blue="$(tty_mkbold 34)"
tty_red="$(tty_mkbold 31)"
tty_reset="$(tty_escape 0)"


# Takes multiple arguments and prints them joined with single spaces in-between
# while escaping any spaces in arguments themselves
shell_join() {
    local arg
    printf "%s" "$1"
    shift
    for arg in "$@"; do
        printf " "
        printf "%s" "${arg// /\ }"
    done
}

# Takes multiple arguments, joins them and prints them in a colored format
ohai() {
  printf "${tty_blue}==> %s${tty_reset}\n" "$(shell_join "$@")"
}

# Takes a single argument, prints it in a colored format and aborts the script
abort() {
    printf "\n${tty_red}%s${tty_reset}\n" "$1"
    exit 1
}

# Check that Aztec versions are set to release versions
read -r -p "Enter the new version number: " VERSION_NUMBER
if [[ -z "$VERSION_NUMBER" ]]; then
    abort "Version number cannot be empty."
fi

# Update build.gradle with the latest version.
test -f "build.gradle" || abort "Error: Could not find build.gradle"
sed -i'.orig' -E "s/ext.gutenbergMobileVersion = '(.*)'/ext.gutenbergMobileVersion = '$VERSION_NUMBER'/" build.gradle || abort "Error: Failed updating gutenbergMobileVersion in build.gradle"

echo "END build.gradle updated";
```

That I then executed on the WordPress-Android Repo. 
Then confirmed that the version in the `build.gradle` file looked as expected.

### DIFF
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -288,12 +288,15 @@ execute "git" "push" "-u" "origin" "HEAD"
 ohai "Create release branch in WordPress-Android"
 execute "git" "switch" "-c" "$WP_APPS_INTEGRATION_BRANCH"
 
+# Explodes the PR url with the second parameter being the PR ID. 
+PULL_ID_ARR=(${GB_MOBILE_PR_URL//pull\// })
+
 ohai "Update build.gradle file with the latest version"
 test -f "build.gradle" || abort "Error: Could not find build.gradle"
-sed -i'.orig' -E "s/ext.gutenbergMobileVersion = '(.*)'/ext.gutenbergMobileVersion = 'v$VERSION_NUMBER'/" build.gradle || abort "Error: Failed updating gutenbergMobileVersion in build.gradle"
+sed -i'.orig' -E "s/ext.gutenbergMobileVersion = '(.*)'/ext.gutenbergMobileVersion = '${PULL_ID_ARR[1]}-$GB_MOBILE_PR_REF'/" build.gradle || abort "Error: Failed updating gutenbergMobileVersion in build.gradle"
 
 execute "git" "add" "build.gradle"
-execute "git" "commit" "-m" "Release script: Update build.gradle gutenbergMobileVersion version to $VERSION_NUMBER"
+execute "git" "commit" "-m" "Release script: Update build.gradle gutenbergMobileVersion to ref"
 
 ohai "Update strings"
 execute "python" "tools/merge_strings_xml.py"
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -1,10 +1,10 @@
 #!/bin/bash
 
 # INSTRUCTIONS FOR TESTING FROM FORKED REPOS
-# Prerequisites: 
+# Prerequisites:
 # 1. Fork the following repos to your github user repo:
 #    a) Gutenberg-Mobile: https://github.com/wordpress-mobile/gutenberg-mobile
-#    * .gitmodules on CURRENT_BRANCH should reference your gutenberg fork, replace 'WordPress' with GITHUB_USERNAME 
+#    * .gitmodules on CURRENT_BRANCH should reference your gutenberg fork, replace 'WordPress' with GITHUB_USERNAME
 #    * (https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/.gitmodules)
 #    b) Gutenberg: https://github.com/WordPress/gutenberg
 #    c) WordPress-Android: https://github.com/wordpress-mobile/WordPress-Android
@@ -22,8 +22,8 @@ WPIOS_TARGET_BRANCH="develop"
 # 4. Update the repo names below to the user repo name for your fork
 GUTENBERG_REPO="WordPress"
 MOBILE_REPO="wordpress-mobile"
-#GUTENBERG_REPO="YOUR_GITHUB_USERNAME"
-#MOBILE_REPO="YOUR_GITHUB_USERNAME"
+# GUTENBERG_REPO="YOUR_GITHUB_USERNAME"
+# MOBILE_REPO="YOUR_GITHUB_USERNAME"
 # 5. Clone the forked gutenberg-mobile repo
 
 # Before creating the release, this script performs the following checks:
@@ -111,13 +111,13 @@ fi
 # Create Git branch
 RELEASE_BRANCH="release/$VERSION_NUMBER"
 ohai "Create Git branch '$RELEASE_BRANCH' in gutenberg-mobile."
-execute "git" "switch" "-c" "$RELEASE_BRANCH" 
+execute "git" "switch" "-c" "$RELEASE_BRANCH"
 
 # Create Git branch in Gutenberg
 GB_RELEASE_BRANCH="rnmobile/release_$VERSION_NUMBER"
 ohai "Create Git branch '$GB_RELEASE_BRANCH' in gutenberg."
 cd gutenberg
-execute "git" "switch" "-c" "$GB_RELEASE_BRANCH" 
+execute "git" "switch" "-c" "$GB_RELEASE_BRANCH"
 cd ..
 
 # Set version numbers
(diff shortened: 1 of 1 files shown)


release_automation.sh
```diff
@@ -287,15 +288,12 @@ execute "git" "push" "-u" "origin" "HEAD"
...
 
-ohai "Update gutenberg-mobile ref"
-cd libs/gutenberg-mobile
-execute "git" "fetch" "--recurse-submodules=no" "origin" "$GB_MOBILE_PR_REF"
-execute "git" "checkout" "$GB_MOBILE_PR_REF"
-execute "git" "submodule" "update"
-cd ../..
+ohai "Update build.gradle file with the latest version"
+test -f "build.gradle" || abort "Error: Could not find build.gradle"
+sed -i'.orig' -E "s/ext.gutenbergMobileVersion = '(.*)'/ext.gutenbergMobileVersion = 'v$VERSION_NUMBER'/" build.gradle || abort "Error: Failed updating gutenbergMobileVersion in build.gradle"
```
mchowning: I don't believe we can set the `vX.XX` version at this time because we're just cutting branches and creating PRs here, we're not actually creating the tagged release (which would result in the vX.XX binary). I think the workflow would go like this.

1. This script creates all the branches and updates the gb-mobile reference in WPAndroid to refer to the binary for the `HEAD` of the gb-mobile release PR ([for example](https://github.com/wordpress-mobile/WordPress-Android/pull/14386/commits/17e279ad89a1a50c578520ed2bf552de1ea43fbe))
2. Each time we update the gb-mobile release branch, we update the WPAndroid gb-mobile binary version to reflect the new `HEAD` of the release branch.
3. We "cut" the gb-mobile release, tagging it with the appropriate version CI picks up the tag and then builds the binary.
4. We manually update the WPAndroid branch to reflect the "tagged" gb-mobile version.

With that said, I haven't actually done this workflow yet myself, so definitely curious to hear any thoughts @jd-alexander  and @guarani  have.

release_automation.sh
```diff
@@ -277,6 +277,7 @@ execute "git" "clone" "-b" "$WPANDROID_TARGET_BRANCH" "--depth=1" "git@github.co
 
 cd "$TEMP_WP_ANDROID_DIRECTORY"
 
+# This is still needed bacause of Andorid Stories.
```
dcalhoun:
```suggestion
# This is still needed because of Android Stories.
```

release_automation.sh
dcalhoun: The points @mchowning laid out make sense to me. Because we tag the `gutenberg-mobile` release _after_ the script creates the `wordpress-android` branch, we cannot rely upon the version number. 

The proposed workflow steps also make sense. Using the tip of `HEAD` for the `gutenberg-mobile` release branch appears to be the approach that [Joel manually used in the previous release](https://github.com/wordpress-mobile/WordPress-Android/blob/17de6cf384659be96dbfa4d974666c5015da3b48/build.gradle#L10) via the `<PR_number>-<full_commit_sha>` pattern. So, if we are able to have the the script recreate this pattern, that likely would work. 

It looks like we "capture" the `gutenberg-mobile` [commit SHA](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/7c82b25d605525bacf6792dcb5b269f3ac2585b5/release_automation.sh#L250) and [PR URL](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/7c82b25d605525bacf6792dcb5b269f3ac2585b5/release_automation.sh#L201) already. We may be able to parse the PR number from the latter. There doesn't appear to be [a `gh` CLI command](https://cli.github.com/manual/gh_pr) for requesting the PR number of the current branch. 

A short term fix could be temporarily removing the `Update gutenberg-mobile ref` section altogether. This would allow us to leverage the toolkit to create the branches and PR, but we'd manually update the `gutenbergMobileVersion`.

dcalhoun (commented):
Thank you for spending time moving this forward! 🙇🏻

release_automation.sh
```diff
@@ -287,15 +288,15 @@ execute "git" "push" "-u" "origin" "HEAD"
...
-execute "git" "commit" "-m" "Release script: Update gutenberg-mobile ref"
+ohai "Update build.gradle file with the latest version"
+test -f "build.gradle" || abort "Error: Could not find build.gradle"
+sed -i'.orig' -E "s/ext.gutenbergMobileVersion = '(.*)'/ext.gutenbergMobileVersion = '${PULL_ID}-$GB_MOBILE_PR_REF'/" build.gradle || abort "Error: Failed updating gutenbergMobileVersion in build.gradle"
+
+execute "git" "add" "build.gradle"
+execute "git" "commit" "-m" "Release script: Update build.gradle gutenbergMobileVersion to ref"
 
 ohai "Update strings"
 execute "python" "tools/merge_strings_xml.py"
```
dcalhoun: It would appear this script was [recently modified](https://github.com/wordpress-mobile/WordPress-Android/commit/03002428012eb459b774927b70448d26d9ae0a43) and now requires [additional arguments](https://github.com/wordpress-mobile/WordPress-Android/blob/a89637138ba84c5576c1d7fb1a8c5c41d244c76a/tools/merge_strings_xml.py#L107-L109). So, we may need to update this command, although I'm not sure what parameters should be passed. However, we may be able to remove this script altogether, if I follow the discussion in p1618416871372200/1618248527.360700-slack-C0126N6JH7Z correctly. WDYT?

dcalhoun (commented):
@enejb I tested running this version of the release script against my [forked repos](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/aa7b4922186244b5014a2606d89bbf136111ac20/release_automation.sh#L3). I noticed two things. 

First, it appears a `build.gradle.orig` file is created in the temporary `WordPress-Android` clone that the release creates. I'm not sure if this is expected or an issue. I just noticed the file listed as an untracked file. 

Second, the script _succeeded_ in updating the `gutenbergMobileVersion` 🎉. However, it then _failed_ during the `tools/merge_strings_xml.py` script. The log message points to the script [exiting early](https://github.com/wordpress-mobile/WordPress-Android/blob/a89637138ba84c5576c1d7fb1a8c5c41d244c76a/tools/merge_strings_xml.py#L107-L109). I have commented further on this issue on the relevant line of code.  

release-script-error

release_automation.sh
mchowning: > we may be able to remove this script altogether

I think this is the right approach because the WPAndroid release process is handling the string updates now.

dcalhoun (approved):
@enejb as discussed with @mchowning in https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/39#discussion_r614028814, I have removed the invocation of the script that generates strings for WPAndroid. After doing so, this script succeeded in its entirety and appeared to include the correct `gutenbergMobileVersion` for WPAndroid. 🎉

I believe we are good to merge this work if you agree with all the changes. Thank you for wrangling this! 🙇🏻 

## Gutenberg Mobile PR
gb-mobile-sha

## WPAndroid PR
wp-android-gb-mobile-version

release_automation.sh
jd-alexander: I agree with the points mentioned above. Allowing the script to update the `gutenbergMobileVersion` with `<PR_number>-<full_commit_sha>` when the branch is created is the appropriate approach as reflected in the latest changes here. Once the gb-mobile release has been merged in we are then able to manually update the `gutenbergMobileVersion` with the gb-mobile release tag of the current release.

enejb:
Just tested this and it worked well for me as well. 
Thank you for all the changes and help to get this across the finish line. 🥳

PR #40, opened 2021-04-16
Add UBE testing to release checklist

cameronvoell:
After discovering this UBE issue: https://github.com/wordpress-mobile/gutenberg-mobile/issues/3349, we realized UBE was left out of existing release test suites, and we decided to add [UBE manual testing](https://github.com/wordpress-mobile/test-cases/blob/trunk/test-cases/gutenberg/unsupported-block-editing.md) to every release. 

I also added some details so that our current testing steps were explicitly added to the P2 checklist and the release PR checklist.

### DIFF
--- a/Releasing.md
+++ b/Releasing.md
@@ -76,6 +76,26 @@ For the body of the post, just copy this checklist and again replace all occurre
 <!-- /wp:paragraph -->
 
 <!-- wp:heading {"level":3} -->
+<h3>Release Testing</h3>
+<!-- /wp:heading -->
+
+<!-- wp:paragraph -->
+<p>o Use the main WP apps to test the new changes in the PR</p>
+<!-- /wp:paragraph -->
+
+<!-- wp:paragraph -->
+<p>o Smoke test the main WP apps for <a href="https://github.com/wordpress-mobile/test-cases/tree/master/test-cases/gutenberg/writing-flow">general writing flow</a></p>
+<!-- /wp:paragraph -->
+
+<!-- wp:paragraph -->
+<p>o Test the Unsupported Block Editor on WP Apps (<a href="https://github.com/wordpress-mobile/test-cases/blob/trunk/test-cases/gutenberg/unsupported-block-editing.md#unsupported-block-editing---test-cases">see steps</a>).</p>
+<!-- /wp:paragraph -->
+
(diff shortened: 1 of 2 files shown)


jd-alexander (approved):
Thanks for these changes to the documentation @cameronvoell All is looking well from my side. Unrelated to these changes, I realized that in the checklist we have Day 1 and then Last day. When I saw that my mind expected to see a Day 2, 3 etc. I guess it could be changed to First Day. I am not even sure if that improves it but I am just thinking out loud 😅 
Also for the Release Testing would we want to specify a day when these get done or it's best if it's just left as a heading?

cameronvoell:
> Thanks for these changes to the documentation @cameronvoell All is looking well from my side. Unrelated to these changes, I realized that in the checklist we have Day 1 and then Last day. When I saw that my mind expected to see a Day 2, 3 etc. I guess it could be changed to First Day. I am not even sure if that improves it but I am just thinking out loud 😅
> Also for the Release Testing would we want to specify a day when these get done or it's best if it's just left as a heading?

Yea I think there is still a balance between making the checklist / release guides highly specific so that the release coordinators have less to think about and consider +saving them some brain cycles versus keeping the release process  flexible and giving release wranglers a little wiggle room for experimentation, and staying engaged with deciding what makes most sense to them. For example I think we recently decided to run an experiment for a few releases where we change the timing of cutting release branches. The update that I made in this case should be applicable for the temporary experiment or the release timing we used previously. 

Maybe after the "experiment" is tested after a few releases we can specify what happens on each day, so I'll leave the Day titles the way they are for now. Thanks for reviewing @jd-alexander !

jd-alexander:
>  For example I think we recently decided to run an experiment for a few releases where we change the timing of cutting release branches. The update that I made in this case should be applicable for the temporary experiment or the release timing we used previously.

Ah, this is such a good point because I was looking at the release experiment and the previous processes and I had realized that the days on which certain tasks are done differ indeed! Thanks for the clarity here!

> Maybe after the "experiment" is tested after a few releases we can specify what happens on each day, so I'll leave the Day titles the way they are for now. 

Sounds like a good plan! 

 > Thanks for reviewing @jd-alexander !

No prob 🙇🏾

PR #42, opened 2021-04-28
Replace Markdown with HTML code tags

dcalhoun:
The checklist pasted into a P2 post must rely upon the HTML `code` tag to be properly formatted by Gutenberg.


cameronvoell (approved):
thanks for the fix!

PR #43, opened 2021-04-28
Add note to notify Excellence Wranglers

dcalhoun:
A direct request was made that we mention the relevant Excellence Wrangler anytime we deliver a beta/hot fix. This will help them remain up-to-date with ongoing releases.


cameronvoell (approved):
Noted the change here. Thanks for documenting 👍

PR #45, opened 2021-05-13
Removed Bintray reference from the release checklist

jd-alexander:
This PR simply updates a reference from Bintray to S3 to avoid any confusion.


fluiddot (approved):
Good catch! LGTM!

PR #46, opened 2021-05-14
Add check localization strings files item to release checklist

fluiddot:
Until we manage to address [the issue related to the localized strings files missing strings and adding extra ones](https://github.com/wordpress-mobile/gutenberg-mobile/issues/3466), we have to add an extra step in the release checklist that requires verifying that these files are generated properly.

Localization strings files:
- [bundle/android/strings.xml](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/bundle/android/strings.xml)
- [bundle/ios/GutenbergNativeTranslations.swift](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/bundle/ios/GutenbergNativeTranslations.swift)

### DIFF
--- a/Releasing.md
+++ b/Releasing.md
@@ -60,6 +60,10 @@ For the body of the post, just copy this checklist and again replace all occurre
 <!-- /wp:paragraph -->
 
 <!-- wp:paragraph -->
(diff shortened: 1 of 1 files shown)


ceyhun (approved):
Thanks!

PR #49, opened 2021-05-29
Correct template for Slack message

guarani:
The Slack template for messaging the main app release team should say that we've merged to `develop`, not to release branches.


dcalhoun (approved):
This makes total sense. I struggled with language to express how the example message should change between scheduled and beta/hot fixes. Thank you for putting this together.

guarani:
> This makes total sense. I struggled with language to express how the example message should change between scheduled and beta/hot fixes. Thank you for putting this together.

Makes sense, it's hard to add conditionals to checklists and I think this change isn't perfect, it just treats scheduled releases as the default and beta/hot fixes as the exception.

PR #52, opened 2021-06-22
Add extra steps to guarantee that the Pods repository is always up-to-date

fluiddot:
This PR adds the following steps to the WordPress iOS PR section:
1. `bundle install`: Installs Ruby dependencies, this is step is required to have the Cocoapods gem up-to-date with the `Gemfile`.
2. `bundle exec pod repo update`: Updates the Pods repository, this will help us to prevent potential errors when installing Pods that have been upgraded recently.

### DIFF
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -343,7 +343,6 @@ ohai "Update GitHub organization and gutenberg-mobile ref"
 test -f "Podfile" || abort "Error: Could not find Podfile"
 sed -i'.orig' -E "s/wordpress-mobile(\/gutenberg-mobile)/$MOBILE_REPO\1/" Podfile || abort "Error: Failed updating GitHub organization in Podfile"
 sed -i'.orig' -E "s/gutenberg :(commit|tag) => '(.*)'/gutenberg :commit => '$GB_MOBILE_PR_REF'/" Podfile || abort "Error: Failed updating gutenberg ref in Podfile"
-execute "bundle" "install"
 execute "bundle" "exec" "pod" "repo" "update"
 execute "rake" "dependencies"
 
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -289,6 +289,7 @@ ohai "Update GitHub organization and gutenberg-mobile ref"
 test -f "Podfile" || abort "Error: Could not find Podfile"
 sed -i'.orig' -E "s/wordpress-mobile(\/gutenberg-mobile)/$MOBILE_REPO\1/" Podfile || abort "Error: Failed updating GitHub organization in Podfile"
 sed -i'.orig' -E "s/gutenberg :(commit|tag) => '(.*)'/gutenberg :commit => '$GB_MOBILE_PR_REF'/" Podfile || abort "Error: Failed updating gutenberg ref in Podfile"
+execute "bundle" "install"
 execute "bundle" "exec" "pod" "repo" "update"
 execute "rake" "dependencies"
 
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -343,8 +343,9 @@ ohai "Update GitHub organization and gutenberg-mobile ref"
 test -f "Podfile" || abort "Error: Could not find Podfile"
 sed -i'.orig' -E "s/wordpress-mobile(\/gutenberg-mobile)/$MOBILE_REPO\1/" Podfile || abort "Error: Failed updating GitHub organization in Podfile"
 sed -i'.orig' -E "s/gutenberg :(commit|tag) => '(.*)'/gutenberg :commit => '$GB_MOBILE_PR_REF'/" Podfile || abort "Error: Failed updating gutenberg ref in Podfile"
+execute "bundle" "install"
+execute "bundle" "exec" "pod" "repo" "update"
 execute "rake" "dependencies"
-
 
 execute "git" "add" "Podfile" "Podfile.lock"
 execute "git" "commit" "-m" "Release script: Update gutenberg-mobile ref"


release_automation.sh
```diff
@@ -343,9 +343,10 @@ ohai "Update GitHub organization and gutenberg-mobile ref"
 test -f "Podfile" || abort "Error: Could not find Podfile"
 sed -i'.orig' -E "s/wordpress-mobile(\/gutenberg-mobile)/$MOBILE_REPO\1/" Podfile || abort "Error: Failed updating GitHub organization in Podfile"
 sed -i'.orig' -E "s/gutenberg :(commit|tag) => '(.*)'/gutenberg :commit => '$GB_MOBILE_PR_REF'/" Podfile || abort "Error: Failed updating gutenberg ref in Podfile"
+execute "bundle" "install"
```
dcalhoun: My understanding is that `bundle install` is executed within `rake dependencies`, and thus leads to the failure when the pod repo is outdated. I believe we should remove this line.

dcalhoun (commented):
I haven't tested this locally yet, but plan to do so later.

release_automation.sh
dcalhoun: Now that I reread this, I believe I am mistaken. I am likely confusing `bundle install` and `pod install`. I often conflate the two tools.

fluiddot: Good catch, I've just verified the [`Rakefile`](https://github.com/wordpress-mobile/WordPress-iOS/blob/d7dad2e5342439d0e83111d6e10c7e33afeab373/Rakefile#L26) and I confirm that it does this step. Although, we're going to run `bundle exec pod repo update` before running `rake dependencies` so the dependencies won't be up-to-date, probably this won't have any side effect, wdyt?

fluiddot: I've just removed the `bundle install` step in [this commit](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/52/commits/dafe490e30fad3e3aa37b994f2d5494034570203).

dcalhoun: Yeah, I'm not sure if `pod repo update` has a dependency on `bundle install` or not. I suppose we can try without it and see what occurs.

fluiddot: Agreed, let's go without it, most likely it won't introduce errors 🤞 .

dcalhoun (commented):
From testing this, the current iteration fails because `bundle install` was not run prior to `bundle exec pod repo update`. 

```
Could not find rake-12.3.3 in any of the sources
Run `bundle install` to install missing gems.

Failed during: bundle exec pod repo update
```

So, we'll either have to revert the last commit or find an alternative approach. I am wondering if it makes more sense to upstream this improvement into [WPiOS' `rake dependencies` task](https://github.com/wordpress-mobile/WordPress-iOS/blob/cabb12d00c0e18098ae14688ec77da86d20842a1/Rakefile#L73-L77). From reading internal messages, I saw someone recommend the following for scripts, to pair ideal speed with a fallback repo update. 

```bash
pod install || pod install --repo-update
```

WDYT?

fluiddot:
> From testing this, the current iteration fails because `bundle install` was not run prior to `bundle exec pod repo update`.
> 
> ```
> Could not find rake-12.3.3 in any of the sources
> Run `bundle install` to install missing gems.
> 
> Failed during: bundle exec pod repo update
> ```

😭 

> So, we'll either have to revert the last commit or find an alternative approach. I am wondering if it makes more sense to upstream this improvement into [WPiOS' `rake dependencies` task](https://github.com/wordpress-mobile/WordPress-iOS/blob/cabb12d00c0e18098ae14688ec77da86d20842a1/Rakefile#L73-L77). From reading internal messages, I saw someone recommend the following for scripts, to pair ideal speed with a fallback repo update.
> 
> ```shell
> pod install || pod install --repo-update
> ```
> 
> WDYT?

I like the idea of doing `pod install || pod install --repo-update` in the [WPiOS' rake dependencies task](https://github.com/wordpress-mobile/WordPress-iOS/blob/cabb12d00c0e18098ae14688ec77da86d20842a1/Rakefile#L73-L77), my only concern is that we might execute `pod install --repo-update` on error cases unrelated to updating the Pod repository. I find out this bash code that might be helpful to replicate in the Rakefile, wdyt?
```
pod install
error_code=$?
if [ $error_code -eq 31 ]; then
    echo "Local specs is not up-to-date, re-running pod install and updating the local spec repo."
    pod install --repo-update
else
    exit $error_code
fi
```

dcalhoun:
> I find out this bash code that might be helpful to replicate in the Rakefile, wdyt?
> 
> ```
> pod install
> error_code=$?
> if [ $error_code -eq 31 ]; then
>     echo "Local specs is not up-to-date, re-running pod install and updating the local spec repo."
>     pod install --repo-update
> else
>     exit $error_code
> fi
> ```

Yeah, this alternative makes sense to me. It appears a bit more robust and might produce more helpful output.

fluiddot:
> > I find out this bash code that might be helpful to replicate in the Rakefile, wdyt?
> > ```
> > pod install
> > error_code=$?
> > if [ $error_code -eq 31 ]; then
> >     echo "Local specs is not up-to-date, re-running pod install and updating the local spec repo."
> >     pod install --repo-update
> > else
> >     exit $error_code
> > fi
> > ```
> 
> Yeah, this alternative makes sense to me. It appears a bit more robust and might produce more helpful output.

Now I've gotta figure out how to do that on Ruby and include it in the `Rakefile` 😄 , I'll let you know when I finish the PR for this 👍 .

release_automation.sh
fluiddot: I'm going to revert the change because I've recently bumped into the same issue when cutting a betafix:
```
==> Create release branch in WordPress-iOS
Switched to a new branch 'gutenberg/integrate_release_1.58.1'
==> Update GitHub organization and gutenberg-mobile ref
Warning: the running version of Bundler (2.2.10) is older than the version that created the lockfile (2.2.21). We suggest you to upgrade to the version that created the lockfile by running `gem install bundler:2.2.21`.
Could not find proper version of rake (12.3.3) in any of the sources
Run `bundle install` to install missing gems.

Failed during: rake dependencies
```

In this case, we need to run `bundle install` before executing `rake`.

fluiddot: Heads up that I created [this PR](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/58) to address the problem with Gem dependencies.

jhnstn:
Closing since this is stale.

PR #53, opened 2021-06-25
Update Release prechecks

geriux:
After upgrading to React Native `0.64` [linking pods are done automatically](https://github.com/WordPress/gutenberg/commit/b4623e3a83ef801f2de3071172fe92ae89305210#diff-9a2250a2cc0b1edbfbdf9cd4edc78323b6bccd9cf02689be56a6b7b90d065d26) so this precheck to avoid including a developer version of `WordPress-Aztec-iOS` is not needed. So I propose to remove this check which currently fails since its not a use case anymore.

I also added a subcase for the `Verify the localization strings files` step of the release checklist.

### DIFF
--- a/Releasing.md
+++ b/Releasing.md
@@ -92,7 +92,7 @@ For the body of the post, just copy this checklist and again replace all occurre
 <!-- /wp:paragraph -->
 
 <!-- wp:paragraph -->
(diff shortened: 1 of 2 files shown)


fluiddot (approved):
LGTM 🎊 !

PR #54, opened 2021-07-09
Create WPAndroid PR last

ceyhun:
WPAndroid CI depends on `Build Android RN Bridge & Publish to S3` task to succeed in gutenberg-mobile PR, and WPAndroid CI always fails right now because WPAndroid PR is created immediately after gutenberg-mobile PR is created and `Build Android RN Bridge & Publish to S3` is still running at that time.

Here we're moving WPiOS PR creation before WPAndroid PR to have a better chance of `Build Android RN Bridge & Publish to S3` in gutenberg-mobile finishing before we create the WPAndroid PR. 

Although WPiOS PR creation takes more time to create than WPAndroid PR, there's still a chance that WPAndroid CI will fail as `Build Android RN Bridge & Publish to S3` task takes ~15min to complete.

### DIFF
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -364,9 +364,7 @@ WP_ANDROID_PR_URL=$(execute "gh" "pr" "create" \
 ohai "WPAndroid PR Created: $WP_ANDROID_PR_URL"
 echo ""
 
-#####
-# Success!
-#####
+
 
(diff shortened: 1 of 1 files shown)


AmandaRiu (approved):
Changes look good. 👍 :shipit:

PR #56, opened 2021-07-15
Update "Release Submission Checklist" in PR template

ceyhun:
Include updating `Unreleased` sections to current version and creating a new `Unreleased` section.


fluiddot (approved):
LGTM 🎊 !

I'm wondering if this will be enough to prevent us from forgetting about updating the release notes and changelog in the future 🤔 . Not sure if it would be overkilling but what do you think about including this in the release checklist template?

ceyhun:
> Not sure if it would be overkilling but what do you think about including this in the release checklist template?

I guess it wouldn't hurt :) I've added it there as well, please take another look @fluiddot.

fluiddot (approved):

PR #58, opened 2021-07-29
Run bundle install before installing iOS dependencies

fluiddot:
I bumped into the following error when executing the release script for betafix 1.58.1:

```
==> Create release branch in WordPress-iOS
Switched to a new branch 'gutenberg/integrate_release_1.58.1'
==> Update GitHub organization and gutenberg-mobile ref
Warning: the running version of Bundler (2.2.10) is older than the version that created the lockfile (2.2.21). We suggest you to upgrade to the version that created the lockfile by running `gem install bundler:2.2.21`.
Could not find proper version of rake (12.3.3) in any of the sources
Run `bundle install` to install missing gems.

Failed during: rake dependencies
```

Looks like we need to assure that the Ruby dependencies are up-to-date before running `rake dependencies`.


SiobhyB (approved):
LGTM, I was able to get the script running through to completion with this addition. 👍

PR #60, opened 2021-09-10
Update readme with some troubleshooting steps

mkevins:
This includes some commands that may be worth trying to troubleshoot if the release script fails.


antonis (approved):
This LGTM 🎉 
Thank you for adding the troubleshooting steps @mkevins 🙇

PR #61, opened 2021-09-22
Clarify day of week regular releases are performed

guarani:
It's my understanding we've been cutting Gutenberg Mobile releases on the Friday (or Thursday) before the main apps cut their release. Previously, we'd cut the Gutenberg Mobile release earlier, on the Monday of the week prior to the main apps cutting their release. 

This just updates the docs to make that clear.

### DIFF
--- a/Releasing.md
+++ b/Releasing.md
@@ -312,7 +312,7 @@ It's best practice to use the automation script (mentioned in the release templa
 
 ### When
 
-On Fridays, the week before the main apps (WPiOS & WPAndroid) have cut their releases, every 2 weeks.
+On Thursdays, the week before the main apps (WPiOS & WPAndroid) have cut their releases, every 2 weeks.
 
 ### Branches
 
--- a/Releasing.md
+++ b/Releasing.md
@@ -312,7 +312,7 @@ It's best practice to use the automation script (mentioned in the release templa
 
 ### When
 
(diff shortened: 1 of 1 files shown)


Releasing.md
```diff
@@ -312,7 +312,7 @@ It's best practice to use the automation script (mentioned in the release templa
 
 ### When
 
-On Mondays, one week before main apps (WPiOS & WPAndroid) have cut their releases, every 2 weeks.
+On Fridays, the week before the main apps (WPiOS & WPAndroid) have cut their releases, every 2 weeks.
```
jd-alexander: I know that we have cut ( and/or finalized) the release on Fridays but since the schedule states that we start on Thursday is it best to put Thursday here instead? Let me know.
release-schedule

jd-alexander: I guess it's fine since the process is started on Thursday and the final integration is done Friday. 
This part of the documentation below says we create and integrate on Friday but it would seem like we _create_ on Thursday and _integrate_ on Friday. Wdyt?
Screenshot 2021-09-27 at 12 49 42 PM

guarani: Sounds good! I wasn't sure if it was Thursday or Friday, but it sounds like Thursday closer matches our processes.

guarani:
Thanks for the review, @jd-alexander! I made the edit and re-requested a review.

jd-alexander (approved):
Looking good @guarani LGTM 🚢

Issue #8, opened 2020-12-08
Allow release_automation.sh to be restarted


guarani:
If aborted, `./release_automation.sh` cannot be re-run safely since it assumes an initial state where none of the steps the scripts performs is already done.
In practice this means that if I abort the script, I have limited options to finish the release:
- Manually clean up any work the script has already done and restart the script from the beginning, or
- Manually perform by hand the remaining steps that were not yet performed by the script

Ideally the script would determine the current release state before executing (e.g. if the gutenberg and gutenberg-mobile steps are already done, resume from the step where WPiOS and WPAndroid PRs are created).

An open question is whether the GitHub API will have all the information the script needs to check the release's current state (e.g. what PRs are open, what branches are made, etc).

fluiddot:
Unfortunately I experienced this in a recent release, in my case I decided to cleanup the work that the script executed.

I think it would be nice to introduce some improvements to make the script re-run safely, here I added some potential improvements:

## Potential improvements
1. Before asking for the new version number, ask if the release process was already run.
2. Before creating the release branches, check if they already exists and if so ask the user to skip the creation and committing version update.
3. Before committing updates to Gutenberg submodule, check if there's any change if not skip it.
4. After generating the bundle, check if there's any change if not skip it.
5. Before creating a PR, check if it already exists and if so ask the user to skip the creation.
6. In WPiOS and WPAndroid steps make something similar about checking the branches and PRs.

dcalhoun:
I agree that adding logic to allow the script to restart wherever it left off is the most ideal outcome. 

One alternative could be adding a `reset` command to the script that merely deletes the relevant local and remote release branches. From my experience, this would at least allow the script to be rerun. I do not think it's the most ideal solution to the problem of script run failures, but, if it's quicker to build a `reset` command, it might be a worthwhile iteration of the toolkit.

fluiddot:
>One alternative could be adding a reset command to the script that merely deletes the relevant local and remote release branches. From my experience, this would at least allow the script to be rerun. I do not think it's the most ideal solution to the problem of script run failures, but, if it's quicker to build a reset command, it might be a worthwhile iteration of the toolkit.

Yeah, I agree it would be quicker to build the reset command than adding conditions to each step of the script to allow it to be restarted. The only concern I have is about the risk of executing the command accidentally and deleting an ongoing release but we could explicitly warn and ask confirmation when it gets executed.

PR #62, opened 2021-09-29
Update the local CocoaPods spec repo

guarani:
Updating the local CocoaPods spec repo helps ensure the release script doesn't fail with a `CocoaPods could not find compatible versions for pod` error.
This error is caused by stale local spec repos and can be fixed with `pod repo update`, so this has been added to the script.

### DIFF
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -66,6 +66,11 @@ gh auth status >/dev/null 2>&1 || abort "Error: You are not logged into any GitH
 # Check that jq is installed
 command -v jq >/dev/null || abort "Error: jq must be installed."
 
+# Updating the local spec repos benefits both Gutenberg and WPiOS which use CocoaPods
+# This avoids the scenario where the script aborts due to a 'CocoaPods could not find compatible versions for pod' error.
+ohai "Update the spec repos located at ~/.cocoapods/repos"
+execute "pod" "repo" "update"
+
 # Check that Aztec versions are set to release versions
 aztec_version_problems="$(check_android_and_ios_aztec_versions)"
 if [[ -n "$aztec_version_problems" ]]; then
@@ -270,9 +275,6 @@ echo ""
 if [[ $REPLY =~ ^[Nn]$ ]]; then
     read -r -p "Enter the branch name you want to target. Make sure a branch with this name already exists in WPiOS repository: " WPIOS_TARGET_BRANCH
 fi
-
-ohai "Update the spec repos located at ~/.cocoapods/repos"
-execute "pod" "repo" "update"
 
 TEMP_WP_IOS_DIRECTORY=$(mktemp -d)
 ohai "Clone WordPress-iOS into '$TEMP_WP_IOS_DIRECTORY'"
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -271,6 +271,9 @@ if [[ $REPLY =~ ^[Nn]$ ]]; then
     read -r -p "Enter the branch name you want to target. Make sure a branch with this name already exists in WPiOS repository: " WPIOS_TARGET_BRANCH
 fi
 
+ohai "Update the spec repos located at ~/.cocoapods/repos"
+execute "pod" "repo" "update"
+
 TEMP_WP_IOS_DIRECTORY=$(mktemp -d)
 ohai "Clone WordPress-iOS into '$TEMP_WP_IOS_DIRECTORY'"
 execute "git" "clone" "-b" "$WPIOS_TARGET_BRANCH" "--depth=1" "git@github.com:$MOBILE_REPO/WordPress-iOS.git" "$TEMP_WP_IOS_DIRECTORY"


jd-alexander:
@guarani I don't see any testing instructions here. Should I run the script locally and verify that no errors occur?

release_automation.sh
```diff
@@ -271,6 +271,9 @@ if [[ $REPLY =~ ^[Nn]$ ]]; then
     read -r -p "Enter the branch name you want to target. Make sure a branch with this name already exists in WPiOS repository: " WPIOS_TARGET_BRANCH
 fi
 
+ohai "Update the spec repos located at ~/.cocoapods/repos"
```
jd-alexander: Would it be better to group this with the other commands starting at line 295? Let me know the rationale for its placement here. Thanks!

guarani: You're right that its placement here doesn't make much sense.
I'm going to move it to the top of the script because it's a command that update all CocoaPod specs, so it's not specific to WPiOS (it also affects Gutenberg).
 I'm going to move it up to the top of the script so that it's one of the first things done and then all other repos that have pods (Gutenberg and WPiOS) can benefit from it.

guarani:
> @guarani I don't see any testing instructions here. Should I run the script locally and verify that no errors occur?

I think the best approach would be to run the script locally, but make the modifications to the script to make it use forks (that must be made of our four repos), more info [here](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/8da7ca1262cdaf3df48373da1b5bc46e249eee09/release_automation.sh#L3).
That said, I tested this by running the script to make the alpha release for smoke testing. Before I ask you to go through the steps to setup the forks, let me just run through those steps myself and I'll re-request a review once I'm sure of the testing steps.

jd-alexander:
> > @guarani I don't see any testing instructions here. Should I run the script locally and verify that no errors occur?
> 
> I think the best approach would be to run the script locally, but make the modifications to the script to make it use forks (that must be made of our four repos), more info [here](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/8da7ca1262cdaf3df48373da1b5bc46e249eee09/release_automation.sh#L3). That said, I tested this by running the script to make the alpha release for smoke testing. Before I ask you to go through the steps to setup the forks, let me just run through those steps myself and I'll re-request a review once I'm sure of the testing steps.

Sounds good to me 👍🏾

fluiddot:
👋 I recall I tried to address this issue in a similar way some time ago in https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/52. However, the PR was paused because we thought that would be a better approach to update the local CocoaPods spec repo in the [WPiOS' rake dependencies task](https://github.com/wordpress-mobile/WordPress-iOS/blob/cabb12d00c0e18098ae14688ec77da86d20842a1/Rakefile#L73-L77) instead of the release script.

enejb:
I would say lets go with this for now solution for now.

The only downside that I see is that it might slow things down. 
Maybe we can have a question like we do for npm install. Where if you are sure you ran this before recently. 
like when you are retrying the script for some reason then you can might want to skip it.

 Other than that I think we should merge it. :)

release_automation.sh
```diff
@@ -66,6 +66,11 @@ gh auth status >/dev/null 2>&1 || abort "Error: You are not logged into any GitH
 # Check that jq is installed
 command -v jq >/dev/null || abort "Error: jq must be installed."
 
+# Updating the local spec repos benefits both Gutenberg and WPiOS which use CocoaPods
+# This avoids the scenario where the script aborts due to a 'CocoaPods could not find compatible versions for pod' error.
+ohai "Update the spec repos located at ~/.cocoapods/repos"
+execute "pod" "repo" "update"
```
fluiddot, line 72: I'm wondering if we should execute this with the `bundle` command to assure that we use the expected `cocoapods` version of WPiOS, wdyt? If so, we should execute this after checking out the WPiOS repository.

mchowning: Any thoughts on ^that^ @guarani ?

mchowning:
Is this something we should also do from within the WPiOS main app directory, or is that unnecessary?

release_automation.sh
guarani: Thanks for flagging, @mchowning. 

👋 @fluiddot to respond to your above question (sorry for not replying earlier): 

> I'm wondering if we should execute this with the bundle command to assure that we use the expected cocoapods version of WPiOS, wdyt?

I think that's valid but I'm not sure there's a right answer. `pod repo update` updates the global `~/.cocoapods/repos` repository, which I understand is shared by all CocoaPods versions wherever they are installed. This suggests to me that `repo update` should be executed using the global `pod` (assuming CocoaPods is installed globally!) rather than a project-specific CocoaPod version (`bundle exec pod ...`) [found in the Gemfile](https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/Gemfile#L4).

Also, while WPiOS uses CocoaPods, so does Gutenberg, so if we don't use the global CocoaPods (via `pod`), it's unclear if we should use [Gutenberg's version](https://github.com/WordPress/gutenberg/blob/ad766590e53a89b2c78ed5cbfe87dde2401699bd/packages/react-native-editor/ios/Gemfile#L3) (1.10.0) or [WPiOS' version](https://github.com/wordpress-mobile/WordPress-iOS/blob/2d53260aae667c67fae1e6f4c1f871a3a839ba02/Gemfile#L4) (1.11) of CocoaPods.

guarani:
> Is this something we should also do from within the WPiOS main app directory, or is that unnecessary?

I don't think it's necessary there. My understanding is that the `CocoaPods could not find compatible versions for pod` error happens when the `~/.cocoapods/repos` directory is stale (i.e. it doesn't have a reference to the version of the Pod that is in the lock file). I think the onus is on the contributor to keep this up-to-date via `pod repo update`.

---

I think this PR might be best closed and left un-merged. The error causes inconvenience to release wranglers because even though `pod repo update` fixes the issue, the changes caused by the script need to be cleaned up before it can be re-ran (https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/issues/8). So fixing that issue might make this error much less impactful and would potentially bring benefits in other scenarios.

Also, if we move the script to CI, I think this error would go away because CI would use a fresh machine that doesn't have an existing (and potentially stale) `~/.cocoapods/repos` folder. This means that on CI, the specs repo would be fetched new from its source and would be up-to-date.

This is my understanding but I might be wrong, so feel free to correct me if you think otherwise.

Any objections to closing this? cc @fluiddot

mchowning:
Perhaps an easier way to handle this would be to just prompt the user when the script first starts running to go to their WordPress-iOS repo and make sure they can run `rake dependencies` on the `trunk` branch before proceeding. It would alert the user to the potential issue early, instead of only doing it toward the end of the script's run, which seems to be the painful part. 

cc: @ttahmouch  in case you have any thoughts here

> Any objections to closing this?

No objection from my end.

dcalhoun:
> Perhaps an easier way to handle this would be to just prompt the user when the script first starts running to go to their WordPress-iOS repo and make sure they can run `rake dependencies` on the `trunk` branch before proceeding. It would alert the user to the potential issue early, instead of only doing it toward the end of the script's run, which seems to be the painful part.

This has become my default practice prior to running the release script — more specifically, merely running `bundle pod repo update`. I agree a warning at the beginning of the script could be a simple solution if automating a fix to this failure is not straightforward or maintainable.

release_automation.sh
fluiddot: Ok, thanks for the explanation 🙇 . I understand that the use of `bundle` is more important when installing the dependencies, as it might change the `COCOAPODS` value in `Podfile.lock`. Being this said, I'll consider my comment solved 👍 .

fluiddot:
> I think this PR might be best closed and left un-merged. The error causes inconvenience to release wranglers because even though `pod repo update` fixes the issue, the changes caused by the script need to be cleaned up before it can be re-ran (#8). So fixing that issue might make this error much less impactful and would potentially bring benefits in other scenarios.
> 
> Also, if we move the script to CI, I think this error would go away because CI would use a fresh machine that doesn't have an existing (and potentially stale) `~/.cocoapods/repos` folder. This means that on CI, the specs repo would be fetched new from its source and would be up-to-date.
> 
> This is my understanding but I might be wrong, so feel free to correct me if you think otherwise.
> 
> Any objections to closing this? cc @fluiddot

No objections from my side, it makes sense to me to close the PR in favor of a different approach, either allowing to retry the command or run it via CI.

> Perhaps an easier way to handle this would be to just prompt the user when the script first starts running to go to their WordPress-iOS repo and make sure they can run rake dependencies on the trunk branch before proceeding. It would alert the user to the potential issue early, instead of only doing it toward the end of the script's run, which seems to be the painful part.

That's a great idea, we currently ask the user to optionally install JS dependencies ([reference](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/41052c2ec2d265e5eae43b4196167ecb97639baf/release_automation.sh#L97-L102)), so we could introduce something similar for the Pods dependencies.

release_automation.sh
guarani: Makes sense, thanks for raising this though 👍

jhnstn:
Closing since this is stale

PR #63, opened 2021-10-04
Move PR comments step to Before the Release section

twstokes:
The step:

```Visit all open gutenberg-mobile PRs that are assigned to X.XX.X milestone and leave a comment with a message similar to the following: ```

is currently under the "Create the Release" section. It seems like this would fit better in the "Before the Release" to add clarification that this should happen before the release cutting process.


guarani (approved):
LGTM! Thanks 👍

PR #64, opened 2021-10-04
Correct step syntax so that all content is shown

twstokes:
The step:

```
Review the release script instructions...
```

currently contains broken syntax that shows an error in Gutenberg.

This PR corrects the syntax so that all of the content is shown and no error appears.

<div>
Before screenshot showing the error and missing content
After screenshot showing the error fixed
</div>


guarani (approved):
Thanks for taking care of this 🙇

PR #68, opened 2021-10-11
Update release timeline.

twstokes:
Updates the Release Timeline to better fit schedule. (P9ugOq-1tO-p2)

### DIFF
--- a/Releasing.md
+++ b/Releasing.md
@@ -10,15 +10,6 @@ npm run bundle
 
 This is useful in case you want to use an unreleased version of the bundle in the apps. For instance, on a PR that's a work in progress, you might want to include to a specific gutenberg-mobile branch in the apps with an updated bundle so reviewers can see the latest changes before approving them (and releasing a new version).
 
-# Release Timeline
-
-Currently, we are experimenting with syncing the Gutenberg Mobile release with the main app release schedule (p9ugOq-1LE-p2). We are also experimenting with performing manual tests every week (p9ugOq-1MA-p2), with a round of "smoke" test before cutting the release and a full round of tests after the release has been integrated into the main apps. A typical Gutenberg Mobile release schedule might look like the following:
-
-- Tuesday of release week: perform a round of "smoke" tests and message all targeted PRs that the release will be cut on Thursday.
-- Thursday of release week: start process to cut the release and integrate it into the main apps. (Main apps are cut upcoming Monday)
-- Tuesday after release week: perform full round of writing flow and sanity tests.
-- Remainder of main app release period: monitor main app release P2 posts for issues found.
-
 # Release Checklist Template
 
 When you are ready to cut a new release, use the following template.
--- a/Releasing.md
+++ b/Releasing.md
@@ -44,12 +44,12 @@ For the body of the post, just copy this checklist and again replace all occurre
 <!-- /wp:paragraph -->
 
 <!-- wp:heading {"level":3} -->
-<h3>Before the Release (Tuesday and Wednesday)</h3>
+<h3>Before the Release (Tuesday)</h3>
 <!-- /wp:heading -->
 
 <!-- wp:group -->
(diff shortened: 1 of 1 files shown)


Releasing.md
```diff
@@ -172,7 +172,7 @@ For the body of the post, just copy this checklist and again replace all occurre
 <!-- /wp:paragraph -->
 
 <!-- wp:heading {"level":3} -->
-<h3>Integrate the Release </h3>
+<h3>Integrate the Release (Friday or earlier)</h3>
```
mchowning: I think we'd be better off just saying Thursday here. Friday should be our "let's make it our goal not to use it but we have it if we need it" cushion imo.

Releasing.md
```diff
@@ -58,11 +58,11 @@ For the body of the post, just copy this checklist and again replace all occurre
 <!-- /wp:group -->
 
 <!-- wp:paragraph -->
-<p>o Midway through the week of the release, create installable builds for WPiOS and WPAndroid based off the current <code>develop</code> branch and complete the <a href="https://github.com/wordpress-mobile/test-cases/tree/master/test-cases/gutenberg/writing-flow">general writing flow test cases</a>. </p>
+<p>o Midway through the week of the release on Wednesday, create installable builds for WPiOS and WPAndroid based off the current <code>develop</code> branch and complete the <a href="https://github.com/wordpress-mobile/test-cases/tree/master/test-cases/gutenberg/writing-flow">general writing flow test cases</a>. </p>
```
mchowning: I think Tuesday is better here so there is some time to fix any issues that are found before we cut the release. Obviously, that opens up more time for "new" problems to sneak in between the tests and the release, so I don't think there is a straightforward answer here.

If we do update to Tuesday, let's also update the heading to just say Tuesday:

> Before the Release (Tuesday and Wednesday)

If we leave this as Wednesday, I think we need to update line 17 to be consistent and also say Wednesday for the smoke tests:

> Tuesday of release week: perform a round of "smoke" tests and...

Releasing.md
```diff
@@ -14,8 +14,8 @@ This is useful in case you want to use an unreleased version of the bundle in th
 
 Currently, we are experimenting with syncing the Gutenberg Mobile release with the main app release schedule (p9ugOq-1LE-p2). We are also experimenting with performing manual tests every week (p9ugOq-1MA-p2), with a round of "smoke" test before cutting the release and a full round of tests after the release has been integrated into the main apps. A typical Gutenberg Mobile release schedule might look like the following:
```
mchowning: This is no longer something we're experimenting with, so I would drop this paragraph, and replace it with something like: "The release checklist contains the details for the release process, but here is a high level timeline..." Alternatively, we could drop the timeline here and instead just refer people to the release checklist so that there is really a single source of truth.

mchowning (commented):
Thanks for these updates Tanner! This is looking much better. I left a couple of minor suggestions, but nothing major.

Releasing.md
twstokes: Sounds good @mchowning!

mchowning:
Also, I updated P9ugOq-1tO-p2 to just reference the `Releasing.md` doc here instead of having a second high-level timeline. Please take a look at that and make any updates there you think might help.

Releasing.md
twstokes: > that opens up more time for "new" problems to sneak in between the tests and the release

Right, agreed.

I'm good with moving the days into a conservative direction! Will do.

Releasing.md
twstokes: Maybe now that we're incorporating days into the checklist we get the best of both worlds and won't have to worry about keeping the checklist in sync with the timeline? I like the single source of truth idea.

mchowning (approved):
Looks great! Thanks Tanner!

guarani:
Thanks for this @twstokes (and @mchowning for reviewing), I missed reviewing it earlier.

Issue #10, opened 2020-12-08
Make script run differently depending on type of release 


guarani:
We could consider tweaking the script to ask what kind of release is needed (ie. a major/minor release or a patch release). That way we might be able to avoid some questions like:

> Do you want to cherry-pick a commit from gutenberg? (y/n)

And make the release process more streamlined and easier for newcomers.

Also, we might wish to "bump the version automatically according to the response as well instead of asking the user to enter the new version" (suggested by @ceyhun <sub>internal reference p9ugOq-1rm#comment-2892</sub>).

dcalhoun:
@ceyhun @Guarani I stumbled upon this from reading documentation. I wanted to share that [vercel/release](https://github.com/vercel/release) uses a similar "What type of release are you creating?" prompt. It also subsequently sets the correct semver tag for you based upon the answer. So, its CLI could be inspiration for whatever we explore next.

Thanks for the work on this release script!

PR #76, opened 2021-10-18
Add instructions on how to create alpha builds

guarani:
Alpha builds are a new type of build for testing. This PR adds guidance here for how to create them, when to use them, and how they differ from other types of releases.

### DIFF
--- a/Releasing.md
+++ b/Releasing.md
@@ -305,12 +305,12 @@ Whenever a build is needed for testing (usually a few days prior to a Regular re
 
 For example, when the next release will be `1.11.0`.
 
-| Repo             | Cut From | Branch Name                               | Merging To      |
-| ---------------- | -------- | ----------------------------------------- | --------------- |
-| gutenberg        | trunk    | rnmobile/release_1.11.0-alpha1            | (do not merge)  |
-| gutenberg-mobile | develop  | release/1.11.0-alpha1                     | (do not merge)  |
-| WPAndroid        | develop  | gutenberg/integrate_release_1.11.0-alpha1 | (do not merge)  |
-| WPiOS            | develop  | gutenberg/integrate_release_1.11.0-alpha1 | (do not merge)  |
+| Repo             | Cut From | Branch Name                               |
+| ---------------- | -------- | ----------------------------------------- |
+| gutenberg        | trunk    | rnmobile/release_1.11.0-alpha1            |
+| gutenberg-mobile | develop  | release/1.11.0-alpha1                     |
+| WPAndroid        | develop  | gutenberg/integrate_release_1.11.0-alpha1 |
+| WPiOS            | develop  | gutenberg/integrate_release_1.11.0-alpha1 |
 
 ### Automation script differences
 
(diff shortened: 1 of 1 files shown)


Releasing.md
```diff
@@ -293,9 +293,36 @@ For the body of the post, just copy this checklist and again replace all occurre
 
 ## Best practices
 
-It's best practice to use the automation script (mentioned in the release template above) for all releases types (regular, betafix, hotfix). When wrangling a betafix or hotfix, it's important to merge the fix to Gutenberg `trunk` independently of the release process. When the release is cut (by the automation script) the commit(s) that make up the betafix or hotfix should then be cherry-picked onto the Gutenberg release branch.
+It's best practice to use the automation script (mentioned in the release template above) for all releases types (alpha, regular, betafix, hotfix). When wrangling a betafix or hotfix, it's important to merge the fix to Gutenberg `trunk` independently of the release process. When the release is cut (by the automation script) the commit(s) that make up the betafix or hotfix should then be cherry-picked onto the Gutenberg release branch.
```
dcalhoun, line 296: I actually have not leveraged the release script when creating alpha release so far. I have manually created PRs to generate installable builds. From reviewing the release before my latest release, it would appear that was the approach for those release wranglers as well. 

I suppose the benefit of leveraging the release script might be avoiding human error and ensuring the Gutenberg ref is the latest. The negative might be a unnecessary branch/PR opened for the `gutenberg` repository.

Releasing.md
```diff
@@ -293,9 +293,36 @@ For the body of the post, just copy this checklist and again replace all occurre
...
+
+### Automation script differences
+
+Compared to a Regular release, the differences here are:
+- When the script asks for the new version number, don't forget to add the `-alpha` suffix (e.g. `1.11.0-alpha1`).
+- All PRs created by the release script should be edited to clarify that they are temporary and will be deleted when testing is finished.
+
+### Release checklist template differences
+
+The release checklist is not used for alpha releases. When testing is finished, please delete all PRs created by the script.
```
dcalhoun:
```suggestion
The release checklist is not used for alpha releases. When testing is finished, please close all PRs and delete all branches created by the script.
```

dcalhoun (approved):
This makes sense to me. While I have not utilized this approach myself, I do not foresee substantial negatives to it as this time. Thanks!

Releasing.md
guarani: I see, I wasn't sure if this was an established approach, but if I recall correctly it was suggested to me and that led to me using it.
Like you say, if this is done manually three PRs are created, but if this is done using the script, four PRs are created. Either way, all must be closed and their branches deleted. Given this, I still think it's worthwhile to use the script since the script creates branch names which will cause optional tests to be triggered on the PRs in WPiOS, WPAndroid, and Gutenberg Mobile. 

With regard to the Gutenberg PR, we could consider tweaking the script to not make a Gutenberg PR for alpha builds, but that might be best left for https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/issues/10 which covers making the script more aware of the type of build it's creating.
Do you think this is a reasonable approach for now?

Releasing.md
guarani: Thanks for the suggestion, added 👍

Releasing.md
dcalhoun: Yes, this sounds like a good path forward to me.

guarani: Thanks @dcalhoun!

PR #82, opened 2021-12-09
Update test plan to only include smoke testing

guarani:
Previously we did extensive testing on the Gutenberg Mobile release before it is merged into the main apps. Nowadays, we only do minimal testing of release to ensure the editor is not broken, then make the release, then do more thorough testing of the editor that's included in the main app beta builds.

This updates the release PR templates to reflect the change.

### DIFF
--- a/release_pull_request.md
+++ b/release_pull_request.md
@@ -27,10 +27,7 @@ No extra PRs yet. 🎉
 
 ## Test plan
 
-- Use the main WP apps to test the changes above.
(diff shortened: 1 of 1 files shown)


guarani:
For clarity, these instructions live on in the release scenario post ([here](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/develop/Releasing.md)), so they will be performed at the right time (after main apps release have been cut).

twstokes (approved):
This makes sense to me @guarani! 🙇

mchowning (approved):

mchowning:
Just giving my second 👍  as a show of support! 😄

PR #84, opened 2022-01-04
Correct the Regular Expression that the stream editor is using to mod…

ttahmouch:
…ify the Gutenberg mobile Version in the `build.gradle` file of WordPress-Android.


enejb (approved):
This looks good to me! Thanks for the clean up!

PR #85, opened 2022-01-14
Add a clean-up step to close the release milestone

guarani:
Adds a clean-up step to close the release milestone. I added at the earliest point in the release after the main app builds have been cut.


guarani:
@geriux, would it make sense to move this to an earlier stage? Like as soon as we cut the Gutenberg Mobile release on Thursday? My thinking is yes, but want to double-check with you.

geriux:
> @geriux, would it make sense to move this to an earlier stage? Like as soon as we cut the Gutenberg Mobile release on Thursday? My thinking is yes, but want to double-check with you.

I agree! I think we could move it, it makes sense to close it around that time 👍

guarani:
Thanks! I've updated the PR and it's again ready for review.

geriux (approved):
LGTM! Thanks for adding this 🚀

guarani:
Thanks for the review!

PR #86, opened 2022-01-17
Add support for multiple cherry-picks

mkevins:
### Description

Often our betafix releases will require that multiple commits are cherry-picked into the Gutenberg branch. This PR enables that from the script so that it doesn't have to be done manually.

#### To test

Run the release script with a bogus version (can append something like `-test` onto the end) and when the prompt comes up for cherry-picking, each response with "Y" should allow a cherry-pick hash to be entered. Then the script should continue asking if more cherry-picking is desired, until the user enters "N" (or something other than "Y" or "y").

After the cherry-picking step, it is enough to `cd` into the `gutenberg` directory to check that indeed the multiple commits were cherry-picked. If any local branches are created during this testing, they can be manually deleted (and it should not be necessary to create any PRs for testing this part of the script).

### DIFF
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -157,12 +157,16 @@ cd ..
 
 # Ask if a cherry-pick is needed before bundling (for example if this is a hotfix release)
 cd gutenberg
-read -r -p "Do you want to cherry-pick a commit from gutenberg? (y/n) " -n 1
-echo ""
-if [[ $REPLY =~ ^[Yy]$ ]]; then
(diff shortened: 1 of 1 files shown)


ttahmouch (commented):
Logically, this looks good to me. However, I haven't attempted to run it myself.

antonis (approved):
Great work @mkevins 👍 
The code looks good and the script worked as expected in my local tests 🎉 
Thank you for improving the release process 🙇

mkevins:
Thanks for reviewing and testing!

PR #87, opened 2022-01-17
Update main apps default target branch to `trunk`

fluiddot:
Recently the `develop` branch in the main apps (WordPress-iOS and WordPress-Android) has been renamed to `trunk`. For this reason, this PR updates the release script to use `trunk` as the default target branch of main apps for the release PRs.

### DIFF
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -17,8 +17,8 @@ WPIOS_PR_LABEL="Gutenberg integration"
 # 3. Ensure that each of your repos contains the target branch listed below:
 GUTENBERG_MOBILE_TARGET_BRANCH="trunk"
 GUTENBERG_TARGET_BRANCH="trunk"
-WPANDROID_TARGET_BRANCH="develop"
-WPIOS_TARGET_BRANCH="develop"
+WPANDROID_TARGET_BRANCH="trunk"
+WPIOS_TARGET_BRANCH="trunk"
 # 4. Update the repo names below to the user repo name for your fork
 GUTENBERG_REPO="WordPress"
 MOBILE_REPO="wordpress-mobile"


geriux (approved):
Thanks for updating this! So nice to see `trunk` in all of them 🤩

fluiddot:
> Thanks for updating this! So nice to see `trunk` in all of them 🤩

Yeah, now we'd only left renaming GB-mobile `develop` branch 😅 .

mkevins:
> now we'd only left renaming GB-mobile `develop` branch

Will this require more than just a rename? IIUC, we may have to change some of the steps in the sequence as well for that :thinking:

fluiddot:
> Will this require more than just a rename? IIUC, we may have to change some of the steps in the sequence as well for that 🤔

I think so, I haven't narrowed down the scripts where the `develop` GB-mobile branch is referenced, but I understand that the change would be very similar to this one.

mkevins:
> I haven't narrowed down the scripts where the `develop` GB-mobile branch is referenced

Afaik, there is no such reference. Instead, the script uses the locally checked out branch for `gutenberg-mobile` and the `gutenberg` submodule. It _does_, however, _target_ the `trunk` branch. I'm thinking that what needs to be done for this is similar to what is currently done for `gutenberg` release branches: create a tag from the `HEAD` of the release branch _prior_ to merging to `trunk`.

Currently, we merge the release branch to `gutenberg-mobile` `trunk`, and _then_ we make a tag (directly from `trunk`). This isn't part of the script, but rather part of the checklist documentation. I think this is what will need to be changed (since we'll be treating `trunk` like our current `develop`, and creating a tag from that _after_ merging would include any work that landed after the code freeze).

cc: @guarani (since I believe you have spent some time thinking about this as well :smiley: ) does this sound correct to you?

guarani:
> I'm thinking that what needs to be done for this is similar to what is currently done for `gutenberg` release branches: create a tag from the `HEAD` of the release branch _prior_ to merging to `trunk`.

I agree with this @mkevins 👍

fluiddot:
> Afaik, there is no such reference. Instead, the script uses the locally checked out branch for gutenberg-mobile and the gutenberg submodule. It does, however, target the trunk branch.

I agree, I think I wasn't clear about mentioning the scripts, I meant that I haven't checked where the `develop` branch is referenced across the project. For example, we would need also to update [the CircleCI configuration](https://github.com/wordpress-mobile/gutenberg-mobile/blob/develop/.circleci/config.yml) as it's also referenced there. However, I did a quick check and apart from the release script and CircleCI, looks like there are no other references.

> I'm thinking that what needs to be done for this is similar to what is currently done for gutenberg release branches: create a tag from the HEAD of the release branch prior to merging to trunk.

Yep, that would be the proper way to tag a new release 👍 .

> Currently, we merge the release branch to gutenberg-mobile trunk, and then we make a tag (directly from trunk). This isn't part of the script, but rather part of the checklist documentation. I think this is what will need to be changed (since we'll be treating trunk like our current develop, and creating a tag from that after merging would include any work that landed after the code freeze).

Good point, with your approach about tagging the `HEAD` of the release branch we would prevent this issue.

I guess most of the work in order to apply this change would be updating the documentation 😅 . What do you think about creating a follow-up issue so we can list all the potential modifications to accomplish it?

mkevins:
> What do you think about creating a follow-up issue so we can list all the potential modifications to accomplish it?

I've created [this issue](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/issues/90) to track this. If you think of / find any other references to `develop` that would need to be updated, please list them as a comment on that issue. :smiley:

PR #89, opened 2022-01-27
Add util `execute_until_succeeds` command and use it for retrying iOS dependency install

fluiddot:
I very often experience errors when running the release command, specifically when installing iOS dependencies via the `rake dependencies` command. For this reason, I decided to add a new util command that retries the passed command until it succeeds (with a maximum of 10 retries count).

In order to test this in isolation:
1. Create a `test.sh` script file.
2. Add the following code:
```
set -e
source ./release_utils.sh

times=0
function fail_after_5_times() {
    echo "This command will fail 5 times until succeeds (params: $@).\n"

    if [[ times -lt 5 ]]; then 
        ((times++))
        return 1
    fi
    echo "Command succeed!"
}

execute_until_succeeds "fail_after_5_times" "param-1" "param-2"
```
3. Run `chmod +x test.sh`.
4. Run `./test.sh`.
5. Observe that the command is retried 5 times until succeeds.


geriux (approved):
LGTM! Thanks for handling this 🚀

PR #91, opened 2022-02-04
Prevent including changes that only affect a single platform in release notes of other platforms

fluiddot:
In relation to the step related to reviewing and updating the release notes of main apps, we don't specify that the changes specific to a platform should only be included in that platform. There have been some cases, in previous releases, where we wrongly introduced changes in WPiOS release notes that only affected Android and the other way around.

In order to prevent this, this PR expands the checklist step, noting that platform-specific changes should only be included in release notes of the platform that affects.


hypest (approved):
LGTM

PR #92, opened 2022-02-09
Add warning about possible WPiOS errors

mchowning:
This is adding a warning and advising anyone running the script to make sure they can get the WPiOS dependencies before proceeding as discussed [here](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/62#issuecomment-1027978731). 

This is just quick and dirty, better than nothing, workaround. We should remove this once we figure out how to make the script more robust so that these errors don't occur (ongoing discussion in https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/62 and https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/52). Until then, it seems better to warn about the error and request a very slight bit of manual work to avoid the risk of the script failing and us having to manually create the WPiOS and WPAndroid PRs.

### DIFF
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -38,7 +38,9 @@ source ./release_utils.sh
 
 # Warn about possible WPiOS errors
 echo ""
-echo "This script will fail when generating the WPiOS PR if your local machine cannot successfully obtain the WPiOS dependencies. For that reason, if you want the script to generate WPiOS PRs, it is STRONGLY recommended that you verify that you can run 'bundle install; rake dependencies' on your local machine from the the WPiOS project's trunk branch before proceeding with the script. Otherwise, the script may fail in the middle of running, and no one wants that."
+echo "This script will fail when generating the WPiOS PR if your local machine cannot successfully obtain the WPiOS dependencies." 
+echo "For that reason, if you want the script to generate WPiOS PRs, it is STRONGLY recommended that you verify that you can run 'bundle install && rake dependencies' on your local machine from the the WPiOS project's trunk branch before proceeding with the script." 
+echo "Otherwise, the script may fail in the middle of running, and no one wants that."
 read -r -p "Are you ready to proceed with the script? (y/n) "
 echo ""
 if ! [[ $REPLY =~ ^[Yy]$ ]]; then
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -39,7 +39,7 @@ source ./release_utils.sh
 # Warn about possible WPiOS errors
 echo ""
 echo "This script will fail when generating the WPiOS PR if your local machine cannot successfully obtain the WPiOS dependencies. For that reason, if you want the script to generate WPiOS PRs, it is STRONGLY recommended that you verify that you can run 'bundle install; rake dependencies' on your local machine from the the WPiOS project's trunk branch before proceeding with the script. Otherwise, the script may fail in the middle of running, and no one wants that."
-read -r -p "Are you ready to proceed with the script? (y/n) " -n 1
+read -r -p "Are you ready to proceed with the script? (y/n) "
 echo ""
 if ! [[ $REPLY =~ ^[Yy]$ ]]; then
     abort "Exiting script..."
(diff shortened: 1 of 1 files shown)


release_automation.sh
```diff
@@ -34,6 +34,19 @@ MOBILE_REPO="wordpress-mobile"
 
 set -e
 
+source ./release_utils.sh
+
+# Warn about possible WPiOS errors
+echo ""
+echo "This script will fail when generating the WPiOS PR if your local machine cannot successfully obtain the WPiOS dependencies." 
+echo "For that reason, if you want the script to generate WPiOS PRs, it is STRONGLY recommended that you verify that you can run 'bundle install && rake dependencies' on your local machine from the the WPiOS project's trunk branch before proceeding with the script."
```
fluiddot, line 44: I think the common failure when generating the WPiOS PR and related to running `rake dependencies` is mainly caused by not having the Cocoapods repository updated (i.e. run the command `pod repo update`). So, I'm wondering if instead of displaying a warning and recommend to run `bundle install && rake dependencies`, we directly ask the user to run `pod repo update`. We could even run the command directly here after the prompt.

Regarding the command `bundle install`, I did a quick check on the `dependencies` Rake task, and looks like the bundles are automatically installed as part of the command ([reference](https://github.com/wordpress-mobile/WordPress-iOS/blob/ce0f67381431e6ffb33a61bd65481cd1e2d51c40/Rakefile#L17-L28)) hence, it shouldn't be necessary to manually run it.

mchowning: > Regarding the command bundle install, I did a quick check on the dependencies Rake task, and looks like the bundles are automatically installed as part of the command ([reference](https://github.com/wordpress-mobile/WordPress-iOS/blob/ce0f67381431e6ffb33a61bd65481cd1e2d51c40/Rakefile#L17-L28)) hence, it shouldn't be necessary to manually run it.

It looks like you're right. I was just trying to exactly echo what the script was going to do here. If we want to drop the separate `bundle install`, I'd rather do that in a separate PR.

> I think the common failure when generating the WPiOS PR and related to running rake dependencies is mainly caused by not having the Cocoapods repository updated (i.e. run the command pod repo update). So, I'm wondering if instead of displaying a warning and recommend to run bundle install && rake dependencies, we directly ask the user to run pod repo update

It seems like if we're just going to ask the user to run `pod repo update` anyway, we might as well do it ourselves and merge https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/62/files. I'm fine with that approach, it just seemed like nothing was happening on these PRs, so I put this simple warning up just to make sure we have something merged to address this. If you'd rather hold off for a more proper fix (either merging #62 or updating the Rakefile) though, I'm fine closing this.

fluiddot: > It looks like you're right. I was just trying to exactly echo what the script was going to do here. If we want to drop the separate bundle install, I'd rather do that in a separate PR.

Ok, that makes sense to me 👍. To be honest, I don't have a strong opinion on whether to display `bundle install` or not. It's true that it would be more accurate to only have `rake dependencies`, but since running `bundle install` twice won't cause any unexpected side effects, we can skip that change if you wish.

> It seems like if we're just going to ask the user to run pod repo update anyway, we might as well do it ourselves and merge https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/62/files. I'm fine with that approach, it just seemed like nothing was happening on these PRs, so I put this simple warning up just to make sure we have something merged to address this. If you'd rather hold off for a more proper fix (either merging https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/62 or updating the Rakefile) though, I'm fine closing this.

I agree, I'd stick with the original purpose of this PR regarding just displaying a warning, so let's omit my comment about giving the option of running the `pod repo update` command 😅 . Besides, as you mentioned, that would be addressed in the PRs you referenced.

fluiddot (approved):
LGTM 🎊 !

As a side note, I also often experienced errors when generating the WPiOS PRs, in fact, I was thinking to update the script to give the option to jump directly to the WPiOS or WPAndroid PR creation steps for those cases.

PR #93, opened 2022-02-09
Check if script is updated

mchowning:
Just adding some checks to check whether the script is unmodified from the remote develop branch. I thought this would be a good idea to make sure someone doesn't accidentally run the script with changes. I made these prompts instead of automatically exiting the script in order to avoid making it too difficult to test changes to the script (of course you could also just comment out these changes while you were testing).

When testing this it may be helpful to modify`${1:-'develop@{upstream}'}` to refer to this PR's branch (so`${1:-'make-sure-script-is-updated@{upstream}'}`.

#### Expected Behavior

| | same commit as remote branch | different commit from remote branch
| --- | --- | ---
| **NO uncomitted changes** | no new prompts | 1 new prompt about not using the HEAD of the remote develop branch
| **uncommited changes** | 1 new prompt about uncommited changes | 2 new prompts: not using HEAD of remote develop and uncommited changes

### DIFF
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -38,14 +38,26 @@ SCRIPT_PATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 cd "$SCRIPT_PATH"
 source ./release_utils.sh
 
+
+# Check if script is up-to-date
 LOCAL_COMMIT=$(git rev-parse HEAD)
 git remote update
 DEVELOP_BRANCH_HEAD=$(git rev-parse ${1:-'develop@{upstream}'})
-
 if ! [[ $LOCAL_COMMIT = $DEVELOP_BRANCH_HEAD ]]; then
     echo ""
     echo "You're not running this script from the HEAD commit on the develop branch." 
-    echo "If you are generating a release you should always use the latest version of the script."
+    echo "If you are generating a release you should generally use the latest version of the script."
+    read -r -p "Are you sure you want the script to proceed? (y/n) "
+    echo ""
+    if ! [[ $REPLY =~ ^[Yy]$ ]]; then
+        abort "Exiting script..."
+    fi
+fi
+
+# Check if script has uncommitted changes
+if [ -n "$(git status --porcelain)" ]; then 
+    echo "You are running this script with uncommitted changes."
+    echo "If you are generating a release you should generally use the current version of the script on the develop branch."
     read -r -p "Are you sure you want the script to proceed? (y/n) "
     echo ""
     if ! [[ $REPLY =~ ^[Yy]$ ]]; then
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -42,8 +42,7 @@ source ./release_utils.sh
 # Check if script is up-to-date
 LOCAL_COMMIT=$(git rev-parse HEAD)
 execute "git" "remote" "update"
-DEVELOP_BRANCH=${1:-'develop@{upstream}'}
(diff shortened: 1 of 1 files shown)


mchowning:
Making this a draft until I figure out how to make shellcheck happy

mchowning:
Shellcheck is now happy, so I'll mark this ready for review.

release_automation.sh
```diff
@@ -34,12 +34,43 @@ MOBILE_REPO="wordpress-mobile"
...
 set -e
 
+SCRIPT_PATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+cd "$SCRIPT_PATH"
+source ./release_utils.sh
+
+
+# Check if script is up-to-date
+LOCAL_COMMIT=$(git rev-parse HEAD)
+git remote update
```
fluiddot: I see that the rest of Git commands executed in the script use the `execute` function, so for consistency, we could use it here too, wdyt?
```suggestion
execute "git" "remote" "update"
```

release_automation.sh
```diff
@@ -34,12 +34,43 @@ MOBILE_REPO="wordpress-mobile"
...
+DEVELOP_BRANCH=${1:-'develop@{upstream}'}
```
fluiddot: I'm wondering in what cases we'd like to provide a different develop branch than `develop` 🤔 , do you foresee a particular situation where we would benefit from having this argument?

fluiddot:
> When testing this it may be helpful to modify${1:-'develop@{upstream}'} to refer to this PR's branch (so${1:-'make-sure-script-is-updated@{upstream}'}.

Another way to test this is by passing the current branch as an argument:
`./release_automation.sh make-sure-script-is-update`

fluiddot (commented):
Just added a couple of comments, apart from that the PR looks good to me 🎊 .
Thanks @mchowning  for introducing this improvement 🙇 .

release_automation.sh
mchowning: 👍  Makes sense. I missed that we were using that.

release_automation.sh
mchowning: My intention wasn't to allow us to provide a different `develop` branch, I basically just want to run `git rev-parse origin/develop` but I used the (super-readable 😉 ) `${1-'develop@{upstream}'}` instead of `origin/develop` to handle if someone had the remote named something other than `origin`. This is the first time I've used this `@{upstream}` syntax, so definitely let me know if you think there's a better way to go about this, I'm not committed to this approach at all.

fluiddot: My comment was more related to the `${1` part of the command. I think what the following command `${1:-'develop@{upstream}'}` is doing is the following ([bash reference](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Parameter-Expansion)):
1. Take the first argument passed to the script as the `DEVELOP_BRANCH` value
2. If it's not provided then it will use the `develop@{upstream}` value.

That's why I asked about the potential situation where we might want to set a different value for `DEVELOP_BRANCH` by an argument (i.e. `./release_automation.sh different-develop-branch`).

mchowning: Ah, I understand now. Thanks for catching that, I wasn't intending to take the branch as an argument. I've pushed an update to remove that.

fluiddot (approved):
LGTM 🎊 !

release_automation.sh
fluiddot: Awesome, thanks for addressing it 🙇 .

PR #94, opened 2022-02-10
Check for release tags instead of release branches

mchowning:
We should be creating our releases off of release tags instead of branches (if we're not using trunk or develop), so updating the script to check for that.

### DIFF
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -28,7 +28,7 @@ MOBILE_REPO="wordpress-mobile"
 
 # Before creating the release, this script performs the following checks:
 # - AztecAndroid and WordPress-Aztec-iOS are set to release versions
-# - Release is being created off of either develop, main, or release/*
+# - Release is being created off of either develop, trunk, or a release tag
 # - Release is being created off of a clean branch
 # - Whether there are any open PRs targeting the milestone for the release
 
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -75,12 +75,17 @@ else
     ohai "Confirmed that Aztec Libraries are set to release versions. Proceeding..."
 fi
 
-## Check current branch is develop, trunk, or release/* branch
+# Check if current HEAD is develop or trunk branch
 CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
-if [[ ! "$CURRENT_BRANCH" =~ ^develop$|^trunk$|^release/.* ]]; then
-    warn "Releases should generally only be based on 'develop', 'trunk', or an earlier release branch."
-    warn "You are currently on the '$CURRENT_BRANCH' branch."
-    confirm_to_proceed "Are you sure you want to create a release branch from the '$CURRENT_BRANCH' branch?"
+if [[ ! "$CURRENT_BRANCH" =~ ^develop$|^trunk$ ]]; then
+
+    # Check if current HEAD is a release tag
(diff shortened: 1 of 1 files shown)


mchowning:
If we do the develop -> trunk switch before this is merged (cc: @jhnstn ) , I should update this PR.

release_automation.sh
```diff
@@ -75,12 +75,17 @@ else
...
-    warn "You are currently on the '$CURRENT_BRANCH' branch."
-    confirm_to_proceed "Are you sure you want to create a release branch from the '$CURRENT_BRANCH' branch?"
+if [[ ! "$CURRENT_BRANCH" =~ ^develop$|^trunk$ ]]; then
+
+    # Check if current HEAD is a release tag
+    CURRENT_TAG=$(git tag --points-at HEAD)
+    if [[ ! "$CURRENT_TAG" =~ v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
+        warn "Releases should generally only be based on 'develop', 'trunk', or a release tag."
+        warn "Gutenberg-Mobile is currently on the '$CURRENT_BRANCH' branch and '$CURRENT_TAG' tag."
+        confirm_to_proceed "Are you sure you want to create a release branch from here?"
```
mchowning, line 87: Maybe it would be worth including a confirmation even if it's a "valid" branch/tag? Something like "You're creating this release off of the ${CURRENT_BRANCH} branch, which has the ${CURRENT_TAG} tag. Do you want to proceed?"

mchowning: Maybe this actually isn't so important since there is also the prompt that says what the current version is and that will alert you if you're not on the tag you intend.

SiobhyB: I agree, I don't think it's so important to duplicate the details of the current branch and tag here as it's already laid out in the warning. In fact, I think the extra details would unnecessarily clutter the messages, which I think are pretty clear as they are:

Screenshot 2022-02-14 at 13 55 35

A nice addition could be to detect when the current tag is an empty string and to then not include a mention of it at all in the warning, but I don't think it's overly important either, as the messaging is still clear.

SiobhyB (approved):
All works as outlined in my testing and looks good to me! The only thing that I think is left is to update one of the comments further up in the file, which references a release branch:

https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/01792865564c49b78c9d3bb2096903cc469ece9d/release_automation.sh#L31

Added my thoughts to your other comment, too. Approving as I don't see any blockers :) 🚀

release_automation.sh
mchowning: I kind of like having the `''` tag as an indicator that there is no tag on the current code. I feel like that could be useful in cases where someone is intending to branch off of a tag but, for example, maybe they checked out a similarly named branch instead. Admittedly, there are ways to make that clearer than just an empty string indicator, but I'm not sure it's worth the effort to make this message more robust. Happy to revisit it if you'd like though.

mchowning:
> the only thing that I think is left is to update one of the comments further up in the file, which references a release branch

Good catch! I updated that comment.

release_automation.sh
SiobhyB: That reasoning makes sense, I agree it's likely not worth the effort to create something more robust when it's clear and understandable as is 👍

PR #100, opened 2022-03-18
Update main apps build instructions on release checklist

jhnstn:
The CI on WPiOS no longer requires the manually kicking off the test build step. 
As such we no longer need to note that step in the release checklist

### DIFF
--- a/Releasing.md
+++ b/Releasing.md
@@ -24,7 +24,7 @@ For the body of the post, just copy this checklist and again replace all occurre
 
 <details><summary>Click to expand</summary>
 <p>
-  
+
 ```html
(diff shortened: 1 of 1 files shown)


Releasing.md
```diff
@@ -24,7 +24,7 @@ For the body of the post, just copy this checklist and again replace all occurre
 
 <details><summary>Click to expand</summary>
 <p>
-
```
jhnstn, line 27: Looks like VSCode removed some tailing whitespace. 
This wasn't intentional but seems fine to keep these changes in this PR

hypest (approved):
LGTM!

PR #102, opened 2022-03-30
Post release template: Removes crossposting to mobilegutenpagesp2

geriux:
This PR removes crossposting to mobilegutenpagesp2.


fluiddot (approved):
Good catch @geriux ! Thanks for updating the template 🙇 .

LGTM 🎊 !

PR #103, opened 2022-03-31
Remove `develop` branch references from Releasing document

fluiddot:
I noticed that we still reference the `develop` branch in the Releasing document, specifically in the `Different types of releases` section. We should remove it since we no longer have a `develop` branch in GB-mobile.


geriux (approved):
LGTM! Thanks for updating it.

PR #104, opened 2022-06-10
Handle the two Pod formats when updating Gutenberg ref in WP-iOS

fluiddot:
Due to the change in the format of Podfile in [this PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/18839/files?show-viewed-files=true&file-filters%5B%5D=#diff-8f7d6adf31268a2d897ee34bd170592648d6e520aa237104395e4a4438af50cbL172), the release script now checks which format is being used for the Gutenberg reference before updating it.

### DIFF
--- a/release_automation.sh
+++ b/release_automation.sh
@@ -336,7 +336,18 @@ execute "git" "switch" "-c" "gutenberg/integrate_release_$VERSION_NUMBER"
 ohai "Update GitHub organization and gutenberg-mobile ref"
 test -f "Podfile" || abort "Error: Could not find Podfile"
 sed -i'.orig' -E "s/wordpress-mobile(\/gutenberg-mobile)/$MOBILE_REPO\1/" Podfile || abort "Error: Failed updating GitHub organization in Podfile"
-sed -i'.orig' -E "s/gutenberg :(commit|tag) => '(.*)'/gutenberg :commit => '$GB_MOBILE_PR_REF'/" Podfile || abort "Error: Failed updating gutenberg ref in Podfile"
+# The Gutenberg reference can be in two different formats in Podfile, hence we have to detect which one is being used before updating it.
+#   - Old format: gutenberg :tag => 'v1.0.0'
+#   - New format: gutenberg tag: 'v1.0.0'
+if ( grep -qE "gutenberg :(commit|tag) => '(.*)'" Podfile ) then
+    # Old format (WordPress-iOS 20.0 and older)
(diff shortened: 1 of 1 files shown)


geriux (approved):
LGTM! I tested it locally and it works well with both versions. Thanks for wrangling this!

We'll need to keep an eye on future release PRs since the `tag:` vs `:commit` change the `:` character position. Since before what I would do is just replace `commit` with `tag` 😅

release_automation.sh
```diff
@@ -336,7 +336,18 @@ execute "git" "switch" "-c" "gutenberg/integrate_release_$VERSION_NUMBER"
...
-sed -i'.orig' -E "s/gutenberg :(commit|tag) => '(.*)'/gutenberg :commit => '$GB_MOBILE_PR_REF'/" Podfile || abort "Error: Failed updating gutenberg ref in Podfile"
+# The Gutenberg reference can be in two different formats in Podfile, hence we have to detect which one is being used before updating it.
+#   - Old format: gutenberg :tag => 'v1.0.0'
+#   - New format: gutenberg tag: 'v1.0.0'
+if ( grep -qE "gutenberg :(commit|tag) => '(.*)'" Podfile ) then
+    # Old format (WordPress-iOS 20.0 and older)
+    sed -i'.orig' -E "s/gutenberg :(commit|tag) => '(.*)'/gutenberg :commit => '$GB_MOBILE_PR_REF'/" Podfile || abort "Error: Failed updating gutenberg ref in Podfile (old format)"
+elif ( grep -qE "gutenberg (commit|tag): '(.*)'" Podfile ) then
+    # New format (WordPress-iOS 20.1 and newer)
+    sed -i'.orig' -E "s/gutenberg (commit|tag): '(.*)'/gutenberg :commit => '$GB_MOBILE_PR_REF'/" Podfile || abort "Error: Failed updating gutenberg ref in Podfile (new format)"
```
fluiddot, line 347: I noticed that we might rather use the new format here, i.e.:
`gutenberg :commit => '$GB_MOBILE_PR_REF'` => `gutenberg commit: '$GB_MOBILE_PR_REF'`

geriux: I tried that and I think it's not compatible 😞 I think it only supports `:commit`

fluiddot: Ah, good catch, thanks for testing it. Let's stick with this option then.

> We'll need to keep an eye on future release PRs since the tag: vs :commit change the : character position. Since before what I would do is just replace commit with tag 😅

Yep, this would be a bit confusing when updating the GB-mobile reference:
- When using commit SHA: `gutenberg :commit => '<SHA>'`
- When using tag: `gutenberg tag: 'vX.Y.Z'`

PR #105, opened 2022-10-20
Add section to README about including reason for beta/hot fixes

fluiddot:
The new section included in this PR aims to provide a better context on the reason a beta/hot fix needs to be included in a version.


derekblank (approved):
LGTM! 🚀

PR #107, opened 2022-12-27
Remove fallback to the old Podfile syntax format

geriux:
This PR updates the logic added in https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/104 to keep compatibility with the old Podfile format, it will now support the new format only from now on.


fluiddot (approved):
Thanks for fixing this @geriux ! LGTM 🎊 !

PR #108, opened 2023-01-05
Releasing - Checklist - Remove step that is no longer necessary

geriux:
This PR removes a step that is no longer necessary, as we don't perform any smoke testing before the actual release day.


fluiddot (approved):
LGTM 🎊 !

PR #112, opened 2023-04-27
Add script to generate the release checklist

jhnstn:
## What
This adds a new script to generate the release checklist.

It will include or exclude steps according to the type of release: scheduled or unscheduled.

Besides the scheduled vs unscheduled variations, the content of the steps are the same as the current release checklist. 
There are however a few minor changes that were made so that the release checklist follows the same pattern as other a8c release checklists:
- The templates use the new P2 task block.
- The checklist generation info, tags, and xposts, are now at the bottom of the list.
- Slight modification to the steps which explicitly mention the release type e.g. removes wording like "if this is a beta/hot fix ..."

## Why
The current checklist: 
- Does not follow the style as all the other a8c app release checklists.
 - Includes steps for all types of releases which can lead to false impressions that a given release type specific step was completed or overlooked.
 - Is embedded in a markdown file which makes it difficult to modify and to incorporate in any future release automation.
 
## How
The script is written in Ruby. I choose Ruby since it's already required to work on the GB mobile project and it ships with ERB support. ERB is a very friendly templating language that should make it easy to update the steps going forward. 

The script includes a helper to render GB P2 task blocks so that the templates can focus on the content of the step and not the Gutenberg syntax.

By default the scripts will fetch the templates from the HEAD of this repository. This way the script has only one dependency ( the script itself ) and that any local changes to the templates are used during execution. There is an environment flag to do local testing (see testing below )  

## Testing

run `./release_checklist.rb -h` to see the usage menu

run `export GBM_USE_LOCAL_RELEASE_TEMPLATES=1` so that the script will use the local templates. Otherwise you will get 404 errors while this PR is un-merged.

### Scheduled releases

- Output to terminal: `./release_checklist.rb 1.0.0` (this can be piped to `pbcopy` if on a mac) 
- Output to a file: `./release_checklist.rb 1.0.0 -o test.html` 
- Try adding a release date with the `-d` flag. The date will only appear in the open milestone issues message.
- Try using the `-m` flag and verify that the release message is **not** added to the output.

- Verify the output looks correct after pasting into a P2 

### Unscheduled releases

Try the same steps as the scheduled release but use a non-zero patch value for the version. ex:
`./release_checklist.rb 1.2.3`

- Verify that the unscheduled release prompt exists the script unless confirmed.

**if not piping to another command:**
- Answer `y` to the add release message prompt
- Your default editor should open ( This has only been confirmed with vim )
- Enter a message, save and exit the editor
- Confirm that the entered message is at the top of the output in a paragraph with the `{"backgroundColor":"pale-cyan-blue"}` block attribute.

- Repeat the above and add a message with the `-m` flag
- Verify that the same blue paragraph appears near the top of the output

**If piping to another command (i.e. `./release_checklist.rb 1.2.3 | pbcopy` ):**
- Verify that the confirm unscheduled release prompt is shown and works as expected.
- Verify that there is no prompt to add a release message interactively 
- Verify that using the `-m` flag does add the release message as expected.

## Example output

The following posts were created by pasting the output from the script 
[Scheduled Release](https://jhnstntestworkinggeneral.wordpress.com/2023/04/27/gutenberg-mobile-1-0-0-release-scenario-2/)
[Unscheduled Release](https://jhnstntestworkinggeneral.wordpress.com/2023/04/27/gutenberg-mobile-1-0-1-release-scenario/)

### DIFF
--- a/release_checklist.rb
+++ b/release_checklist.rb
@@ -142,7 +142,7 @@ option_parser = OptionParser.new do |opts|
     options[:release_date] = d
   end
 
-  opts.on '-v', '--mobile-version', 'Mobile host version' do |v|
+  opts.on '-v', '--mobile-version', 'Mobile host version (Only used for unscheduled releases)' do |v|
     options[:mobile_version] = v
   end
 
--- a/release_checklist.rb
+++ b/release_checklist.rb
@@ -64,7 +64,7 @@ class ReleaseChecklist < Checklist
   def initialize(version, options = {})
     @version = version
     local_template = options.fetch(:template, false)
-    @template = local_template ? File.read(local_template) : self.class.template
+    @template = self.class.template
     @release_date = options.fetch(:release_date, "[date]")
     @mobile_version = options.fetch(:mobile_version, "XX.X")
     @is_unscheduled = options.fetch(:is_unscheduled, false)
@@ -144,10 +144,6 @@ option_parser = OptionParser.new do |opts|
 
   opts.on '-v', '--mobile-version', 'Mobile host version (Only used for unscheduled releases)' do |v|
     options[:mobile_version] = v
-  end
-
-  opts.on '-t', '--template TEMPLATE', 'Template file' do |t|
-    options[:template] = t
   end
 
   opts.on '-o', '--output OUTPUT', 'Output file' do |o|
--- a/release_checklist.rb
+++ b/release_checklist.rb
@@ -1,0 +1,222 @@
+#!/usr/bin/env ruby
+require 'erb'
+require 'open-uri'
+require 'optparse'
+require 'tempfile'
+
+def use_local_template?
+  ENV['GBM_USE_LOCAL_TEMPLATES']
+end
+
+# Output helpers
+def say(message)
+  STDERR.print(message)
+end
+
+def exit_with(message)
+  say(message)
+  exit
+end
+
+def abort_with(message)
+  say(message)
+  exit(1)
+end
+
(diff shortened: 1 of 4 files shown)


release_checklist.rb
```diff
@@ -0,0 +1,218 @@
...
+    super(@template)
+  end
+
+  def result
+    super(binding)
+  end
+
+end
+
+class AztecChecklist < Checklist
```
jhnstn, line 86: Breaking out the Aztec checklist into it's own template is that we could extract the "verify Aztec" checks (some variation of [this](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/trunk/release_prechecks.sh#L33) ) and only include this part of the checklist if needed.
As is we this part of the checklist is hardly used.

release_checklist.rb
```diff
@@ -0,0 +1,218 @@
...
+  def result
+    super(binding)
+  end
+
+  def self.render()
+    new().result
+  end
+end
+
+class IncomingChangeChecklist < Checklist
```
jhnstn, line 112: Similar to the Aztec checklist, I separated out the incoming changes so that we can only include these steps when needed. In this case we could update the checklist generator script to output just the "incoming changes" and then the release wrangler appends the block of steps to the existing checklist.

I should also mention that the release window has narrowed so that I'm not entirely sure we need these steps at all. We now prepare and cut the release the same day. Even if we need to make a last minute change I not sure it warrants a full set of checklist steps.

PR #140, opened 2023-06-29
fix: iOS update ignores trailing comma

dcalhoun:
Ignoring the trailing comma makes this script more resilient.

Including the trailing comma in the matcher, but not the replacement,
resulted in removing the trailing comma. This meant the next run would
fail as the trailing comma would no longer be present. Also, the
trailing comma may be removed by chance when the commit is manually
updated during normal development, which can occur and is quite
harmless.

The original matcher was created in dc44c49. Examples of the trailing
comment changing can be found in the following:

- https://github.com/wordpress-mobile/WordPress-iOS/pull/20827/commits/b7b34c885d44b1bfe483d18f54d564970748115c
- https://github.com/wordpress-mobile/WordPress-iOS/commit/f8646c968c210cf888601985b2e01f0d07db519a


mokagio (approved):
This will get better soon once we move this config [to a YAML file](https://href.li/?https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/114#issuecomment-1543385142).

PR #145, opened 2023-09-14
Update Aztec-iOS podspec source path to point Gutenberg

fluiddot:
This is a required change after https://github.com/wordpress-mobile/gutenberg-mobile/pull/6200 was introduced into Gutenberg Mobile. Now we only use the Aztec-iOS podspec located in Gutenberg, so we need to update the path used when checking the version.


mokagio (approved):
Looking good. I didn't run the script but I copied the new path and `cat` it locally, just to make sure no typo slipped through.

PR #155, opened 2023-10-09
[CLI] Add `release prepare` command

jhnstn:
Adds the `release prepare` command.

The command prepares a Gutenberg release PR.

## Testing
- Set`GBM_WORDPRESS_ORG` to a forked version of Gutenberg
    - This can be done inline with the command i.e. `GBM_WORDPRESS_ORG=your_org go run main.go release prepare v1.0.0` or by setting the env in your shell config i.e. `export GBM_WORDPRESS_ORG="your_org"`. The later is preferred if working on a lot of changes)
- run `go run main.org release prepare {version}` and answer the prompts to create a release pr on the forked repo


cli/go.mod
```diff
@@ -1,8 +1,11 @@
 module github.com/wordpress-mobile/gbm-cli
 
-go 1.20
+go 1.21
```
derekblank, line 3: I was prompted to add these changes to `go.mod` after running `go run main.go release prepare v1.0.0`, which asked me to run `go mod tidy`. I think it may be because I was using Go `v1.21`, when the go.mod file was set on Go `v1.20`. These dependency updates seem fine, but let me know if there's something to update here -- or perhaps there's a way to enforce a common Go version for the project, similar to `nvm` checks? Overall I think it's a minor point, but here's the command output I was following that prompted this change:

```
release-toolkit-gutenberg-mobile/cli on  cli/add-prepare-cmd [$] via 🐹 v1.21.2 
➜ GBM_WORDPRESS_ORG=derekblank go run main.go release prepare v1.0.0
go: updates to go.mod needed; to update it:
        go mod tidy

release-toolkit-gutenberg-mobile/cli on  cli/add-prepare-cmd [$] via 🐹 v1.21.2 
➜ go mod tidy
go: downloading github.com/stretchr/testify v1.7.0
go: downloading gopkg.in/h2non/gock.v1 v1.1.2
go: downloading github.com/inconshreveable/mousetrap v1.1.0
go: downloading golang.org/x/image v0.6.0
go: downloading golang.org/x/mobile v0.0.0-20230301163155-e0f57694e12c
go: downloading gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127
go: downloading github.com/davecgh/go-spew v1.1.1
go: downloading github.com/pmezard/go-difflib v1.0.0
go: downloading github.com/kr/pretty v0.1.0
go: downloading golang.org/x/exp v0.0.0-20190731235908-ec7cb31e5a56
go: downloading github.com/h2non/parth v0.0.0-20190131123155-b4df798d6542
go: finding module for package github.com/kr/text
go: downloading github.com/kr/text v0.2.0
go: found github.com/kr/text in github.com/kr/text v0.2.0
```

cli/pkg/console/console.go
```diff
@@ -59,7 +59,8 @@ func Log(format string, args ...interface{}) {
 }
 
 func Debug(format string, args ...interface{}) {
-	l.Printf(format+"\n", args...)
+	blue := color.New(color.FgBlue).SprintfFunc()
+	l.Printf(blue("\n"+format, args...))
```
derekblank, line 71: I experimented with updating Error and Debug colors here just to get some Go-fu. We don't have to stick with these actual color values in the long-run, but they appear to be working fine. 👍 Go is fun.

cli/pkg/console/console.go
```diff
@@ -1,20 +1,34 @@
 package console
 
 import (
+	"bufio"
 	"fmt"
+	log "log"
```
derekblank: I was curious why `log "log"` was duplicated here, particularly if it's a quirk to Go that modifies the behavior of `log` here. I follow that capital-letter declarations are public and lowercase are private, but not sure if that rule applies here (especially since the import and the alias are both lowercase). Not at all a big deal, but just wanted to learn more about what this declaration does.

cli/pkg/exec/exec.go
```diff
@@ -0,0 +1,55 @@
...
+
+func SetupNode(dir string, verbose bool) error {
+	// Check for nvm
+	_, ok := os.LookupEnv("NVM_DIR")
+	if ok {
+		exc(verbose, dir, "nvm", "use")
+	}
+
+	return nil
+}
```
derekblank, line 30: This function gave me some problems when running the `prepare` command initially. When running the command, I kept failing at this step with errors about the Node version: `npm ERR! code EBADENGINE`. 

We discussed this already, but we may need to expand how we handle Node environments, especially for anyone who isn't using `nvm`. Added an issue for this here: https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/issues/157.

cli/cmd/release/prepare.go
```diff
@@ -0,0 +1,36 @@
...
+		console.ExitIfError(err)
+
+		defer utils.CleanupTempDir(tempDir)
+
+		console.Info("Created temporary directory %s", tempDir)
+
+		pr, err := release.CreateGbPR(version, tempDir)
+		console.ExitIfError(err)
+
+		console.Info("Created PR  (url ? %s)", pr)
```
derekblank: The output of this command seemed like it might not be quite right when I ran this. Some values from the API appear to be missing, and the formatting does not seem like it's using the correct line breaks. I think this may be because I did not set up the API auth token correctly for my own personal GitHub organization when testing, which results in some of the missing or nil values in the response. Not at all a blocker to merging this PR, but just posting this to see if your output here from this command is any different. Here's the text of the terminal output (and a screenshot from the terminal to check the syntax highlighting, if that helps):

```
To github.com:derekblank/gutenberg
 * [new branch]      rnmobile/release_1.0.1 -> rnmobile/release_1.0.1
PR created: &{%!!(MISSING)s(int=2) https://github.com/derekblank/gutenberg/pull/2 https://api.github.com/repos/derekblank/gutenberg/pulls/2 ## Description
Release 1.0.1 of the react-native-editor and Gutenberg-Mobile.

 Mobile Release v1.0.1 [{Mobile App - i.e. Android or iOS}] open {derekblank} %!!(MISSING)s(bool=false) %!!(MISSING)s(bool=false) {rnmobile/release_1.0.1 f755a74fe3b45194f73d95a724bb9a89924fde98 {}} {trunk 416bab68238be99fa77429d9d6a1e0174fdd9507 {}} []  }

Created PR  (url ? {%!!(MISSING)s(int=2) https://github.com/derekblank/gutenberg/pull/2 https://api.github.com/repos/derekblank/gutenberg/pulls/2 ## Description
Release 1.0.1 of the react-native-editor and Gutenberg-Mobile.

 Mobile Release v1.0.1 [{Mobile App - i.e. Android or iOS}] open {derekblank} %!!(MISSING)s(bool=false) %!!(MISSING)s(bool=false) {rnmobile/release_1.0.1 f755a74fe3b45194f73d95a724bb9a89924fde98 {}} {trunk 416bab68238be99fa77429d9d6a1e0174fdd9507 {}} []  })
```

Screenshot 2023-10-10 at 4 32 12 pm

derekblank (approved):
This is looking great. I'm able to successfully run commands like `GBM_WORDPRESS_ORG=derekblank go run main.go release prepare v1.0.1` and review the output. 

I made some minor code updates based on some of my explorations today, and added some review comments as well. Otherwise, I think this is good to merge. 👍

cli/pkg/exec/exec.go
derekblank: Also noting that I was only able to get around an exit here by setting my Node global default, e.g., `nvm alias default v16.20.2`. It's totally fine for now to keep things moving, but we could aim to set the Node version within this command, independent of the user's global default version.

cli/pkg/console/console.go
jhnstn: Oh, I think I left that in when I was showing package aliases.  
Here is the go docs on package naming https://go.dev/blog/package-names. I'm not sure if it's just a style thing but package names are lowercase, even though they are "public". It could be the case that if the alias is capitalized then someone could access the `log` package via `console.Log`. Can't think of a great reason to do that however.

cli/pkg/exec/exec.go
jhnstn: Totally agree that this should be more robust.
 I was assuming folks would have the required node version installed to work on the project and that `nvm use` would be able to pick that up. 
On the other hand I might have had to tweak my machine to get this working back in the day.
Thanks for adding the issue!

cli/go.mod
jhnstn: Hmm not sure what to do here. I did notice that the `go.mod` file no longer works with go v1.20. It was complaining about the `toolchain` directive.

I updated my go version to 1.21.3 and it seems to be working again. Doing this just to unblock the work but we should try to figure out a better way to manage the go version.

jhnstn: here are the docs on `toolchain` for future reference https://go.dev/doc/toolchain

jhnstn: So this kinda messed up my vscode setup.  I was unable to get vscode to recognize go v1.21.3, even after removing v1.20 from my machine. 
I had to remove `GOROOT` from my zsh config per the advice in this SO post https://stackoverflow.com/a/76321397/1373043

cli/cmd/release/prepare.go
jhnstn: It looks like the [PR was created ](https://github.com/derekblank/gutenberg/pull/2)but it didn't add labels 🤔 
That might be a GH auth issue but seems odd that you could only create the PR but not the labels.

I think the odd output has more to do with printing the `gh.PullRequest`. The gh api request places the response in the `gh.PullRequest` struct but I wonder if those missing values relate to properties in the response interface which are not in the struct.

jhnstn: Here is a bit more on the `MISSING` stuff https://www.reddit.com/r/golang/comments/j7ryfd/why_does_a_missing_argument_or_type_mismatch_in/

jhnstn: Just realized why the labels were not created.
The labels have to exist on the repo before adding them to a PR. 
I'm open on how to handle this scenario. The command should least make it clear that the labels were not present.  The issue is that with the current GH api we have to add the labels after we create the PR, so if we want to fail the command we should first check if the label exists before creating the PR.

Given that the labels are hard coded to match the main repos I think it's a bit overkill to do the label check. I think just making it very clear warning that the labels failed should be enough. Perhaps warn with "Adding the labels failed, are you sure the label {label} exists on the repo?"

jhnstn:
@derekblank I added a few more changes. I'm going to merge so we can move on to other work, Thanks for the review!

cli/cmd/release/prepare.go
derekblank: > The labels have to exist on the repo before adding them to a PR.

That makes sense! 

> Given that the labels are hard coded to match the main repos I think it's a bit overkill to do the label check. I think just making it very clear warning that the labels failed should be enough.

I think this is plenty. The labels will already exist on the GB and GBM repos, so the chance of encountering this would be very minimal in production. Plus, it's trivial to add the labels manually.

cli/go.mod
derekblank: We can revert this commit if we need to, and stick with Go `v1.20`. It wouldn't let me _not_ run `go mod tidy` on `v1.21`, however.

Issue #152, opened 2023-10-06
Add command to prepare GBM release


jhnstn:
The command should
- Update the GB reference to the release tag *
- Prepare the bundle 
- Update the release notes
- Create the release PR

* This is based on the proposal that the prepare GB step will add the release tag.

jhnstn:
What should we do if this command is ran twice? 

It should be reenterant so that if it fails it can be ran again with out causing unexpected behavior i.e. a failed run followed by a successful run should have the same result as a single successful run.

But what should we do if ran after a successful run?
Should it attempt to update or fail? 
In the former case we could add an update flag or have a completely separate update command.

The only redundant work in an update scenario is the PR creation. We should account for this regardless of how we approach updating

derekblank:
* https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/158

PR #158, opened 2023-10-11
Add command to prepare GBM release

derekblank:
Adds `prepare` function for Gutenberg Mobile:

* https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/issues/152

The intent of the Gutenberg Mobile prepare command:

- Update the GB reference to the release tag
- Prepare the bundle
- Update the release notes
- Create the release PR


derekblank:
@jhnstn This is currently a WIP based on the previous version of [gbm-cli/cli/pkg/gbm/gbm.go](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/gbm-cli/cli/pkg/gbm/gbm.go).

There are still a few items to address and get feedback on:


- [x] The structure of the original `gbm-cli` branch and the branch the current `trunk` (with the Gutenberg `prepare` command) are a bit different. Does the directory structure proposed here make sense, or should we move where we want to place this Gutenberg Mobile prepare command?

- [x] In gbm.go in this PR, I started updating some of the git methods from using `go-git` to using our own `git.go` methods. We already discussed not using `go-git` library due to some of its limitations (which I agree is a good idea), but wanted to check on the methods that this PR adds in `git.go`. Some of them still need some updating, particularly with the pointers.

cli/pkg/git/git.go
```diff
@@ -14,6 +16,30 @@ func Clone(repo, dir string, shallow bool) error {
 	return cmd("clone", repo)
 }
 
+func CloneGBM(dir string, pr gh.PullRequest, verbose bool) (*g.Repository, error) {
```
jhnstn: I'm not sure about putting something this GBM specific in the `git` package.  Feels like it belongs in `gbm/gbm.go`

Not a blocker but something to consider as a refactor.

cli/pkg/git/git.go
```diff
@@ -14,6 +16,30 @@ func Clone(repo, dir string, shallow bool) error {
...
+	git := exec.ExecGit(dir, verbose)
```
jhnstn: I was thinking of doing something like this at the top level of the package.
Here are the proposed changes https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/161
The gist is we can remove having to supply the directory verbose setting for every git command. I also set it up so it can take any of the underlining git command flags/args.

cli/pkg/git/git.go
```diff
@@ -14,6 +16,30 @@ func Clone(repo, dir string, shallow bool) error {
...
+
+	org, _ := repo.GetOrg("gutenberg-mobile")
+	url := fmt.Sprintf("git@github.com:%s/%s.git", org, "gutenberg-mobile")
+
+	cmd := []string{"clone", "--recurse-submodules", "--depth", "1"}
+
+	fmt.Println("Checking remote branch...")
+	// check to see if the remote branch exists
+	if err := git("ls-remote", "--exit-code", "--heads", url, pr.Head.Ref); err != nil {
```
jhnstn: I _think_ some of this is in here to handle cases where we need to re-run the command. In theory the branch shouldn't be there. 

I know that eventually we'll want to handle cases where the branch exists, like in an update or publish scenario but I think we have more to do to handle those cases.

I think for now it's fair to assume the release branch does not exist.

cli/pkg/git/git.go
```diff
@@ -29,3 +55,58 @@ func CommitAll(dir, format string, args ...interface{}) error {
...
+	}
+
+	if err := git("commit", "-m", message); err != nil {
+		return fmt.Errorf("unable to commit submodule update %s : %s", submodule, err)
+	}
+	return nil
+}
+
+func IsSubmoduleCurrent(s gh.Submodule, expectedHash string) (bool, error) {
+	// Check if the submodule is porcelain
```
jhnstn: I know this is still WIP but looks like most of this is from the go-git package. It should be easier to use exec.Git here

cli/pkg/gbm/gbm.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil, err
+	}
+
+	return gbmr, nil
+}
+
+func setupGb(gbmDir string, gbmr gh.Repo, gbPr *gh.PullRequest, verbose bool) error {
+	console.Info("Checking Gutenberg")
+
+	gb, err := git.GetSubmodule(gbmr, "gutenberg")
```
jhnstn: We should have to update all the submodules since this will be a fresh clone. We can do this before the `setupGb` call above.

cli/pkg/gbm/gbm.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return err
+	}
+	if curr, err := git.IsSubmoduleCurrent(gb, gbPr.Head.Sha); err != nil {
```
jhnstn: We shouldn't need all this checking if we are working off a fresh clone. I might have added this stuff in my original cli code but I don't think it's needed.

I think all we really need to do is cd into gutenberg and switch to the gb release branch. I don't even think we need to do any npm stuff (but I do think that will happen due to how we have npm install set up on gbm )

cli/pkg/release/gbm.go
```diff
@@ -0,0 +1,148 @@
...
+	console.Info("Checking if branch %s exists", branch)
+	exists, _ := gh.SearchBranch("gutenberg-mobile", branch)
+
+	if (exists != gh.Branch{}) {
+		console.Info("Branch %s already exists", branch)
+		return pr, nil
+	} else {
+		console.Info("Cloning Gutenberg Mobile to %s", dir)
+
+		err := git.Clone(repo.GetRepoPath("gutenberg-mobile"), "--depth=1")
```
jhnstn: @derekblank  I made a similar change in `gb.go`
```suggestion
		err := git.Clone(repo.GetRepoPath("gutenberg-mobile"), "--depth=1", ".")
```
The problem with letting `git.Clone` cloning into the default directory (`gutenberg-mobile`) is that the git client is configured to the parent directory `dir`.

With this change we'll also have to replace  `gbmDir` on line 19 with `dir`

jhnstn: Or the other option is to create a separate git client for cloning and another for all the other git operations. I thought cloning into `.` was cleaner for gb.go.

derekblank:
@jhnstn Updated the PR with the output of our pairing session and then continued the rest of the steps to create the PR based on (gb.go and the previous release-automation script). There may be some tweaks needed to gbm.go as you work it into a release command, but otherwise marking this release pkg PR as read to review. Feel free to make any tweaks on this branch as well, or we can merge and iterate from there to keep code consolidated more quickly.

cli/pkg/release/gbm.go
```diff
@@ -0,0 +1,197 @@
...
+	if err := git.CommitAll(gbmDir, "Release script: Update gutenberg submodule"); err != nil {
+		return pr, err
+	}
+
+	// Run npm ci and npm run bundle
+	if err := exec.NpmCi(gbmDir, true); err != nil {
+		return pr, err
+	}
+
+	if err := exec.NpmRun(gbmDir, true, "bundle"); err != nil {
```
jhnstn: So I just ran into this today on release `1.106.0` where there were no changes to commit after bundling. This was new to me but all really bundle for these days are the i18n strings. Turns out that there are no new strings for this release and the `get commit ...` command failed.

We should check to see if the repo is porcelain at this point before we try committing.
We could add that check to `CommitAll` and send a warning message if there is nothing to commit. 
Might be overloading the commit operation however.

derekblank:
@jhnstn I merged the `prepare` subcommands from trunk, and wired up the GBM prepare command to call `CreateGbmPR`, e.g.:

```
GBM_WORDPRESS_ORG=derekblank go run main.go release prepare gbm v1.0.6
```

It's invoking the `CreateGbmPR` command correctly, but there's still an issue with the tmp directory. Perhaps it's how we're passing tempDir [here](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/ba0a45730485be0e41052b139cbdcdd20b52f951/cli/cmd/release/prepare/gbm.go#L34), or setting the `gbmDir` [here](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/ba0a45730485be0e41052b139cbdcdd20b52f951/cli/pkg/release/gbm.go#L20).

derekblank:
@jhnstn I believe the issue with `nothing to commit, working tree clean` was that the remote origin and upstream branch was not being set in git. I added git interfaces to add these to the GBM prepare command, but the syntax might need some tweaks to get it right.

jhnstn:
@derekblank I got this working. The PR body is empty but going to merge as is

jhnstn (approved):

Issue #156, opened 2023-10-10
CLI: Checklist displays the wrong release date


jhnstn:
The checklist generator shows a suggested message to add to open milestone issues.
It should calculate date for the Thursday following the day the checklist was rendered. 

Currently it's inaccurate. 
The following screenshots were taken after running the script on October 9th. Note that the correct date should be October 12th but was rendered as October 10th.

## Expected
<img width="770" alt="Screenshot 2023-10-09 at 9 22 22 PM" src="https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/assets/744755/e4329eb7-10ce-4ec6-bac7-af42ec0e3679">

## Actual 
<img width="751" alt="Screenshot 2023-10-09 at 9 22 07 PM" src="https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/assets/744755/ad18c123-6d88-4e61-8d3c-4732543873bd">

derekblank:
This would be a good one for me to take.

PR #159, opened 2023-10-11
Update NextReleaseDate util to display the correct date

derekblank:
Updates the CLI `NextReleaseDate` util to display the correct next release date:

* https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/issues/156

**To test:** 

Run `go run main.go render checklist -v 1.0.0` from the `cli` directory, and verify that the date output in the `Hey [author]` quote is the correct date for next Thursday:

```
Hey [author]. We will cut the 1.0.0 release on Thursday October 12, 2023. I plan to circle back and bump this PR...
```

I believe the original issue was because the first Monday in January 2006 was January the 2nd, not January the 1st:

```
- return nextThursday.Format("Monday 01, 2006")
+ return nextThursday.Format("Monday January 2, 2006")
```

I also added the month, which I though should be included. Previously, the command would render "We will cut the release on Thursday 10, 2023", which felt a bit odd without the name of month included. Also updated the test to reflect this.

### DIFF
--- a/cli/pkg/utils/utils.go
+++ b/cli/pkg/utils/utils.go
@@ -25,7 +25,7 @@ func NextReleaseDate() string {
 
 	nextThursday := time.Now().AddDate(0, 0, int(daysUntilThursday))
 
-	return nextThursday.Format("Monday 01, 2006")
+	return nextThursday.Format("Monday January 2, 2006")
 }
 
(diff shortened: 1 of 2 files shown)


jhnstn (approved):
Nice catch on the format!

Issue #163, opened 2023-10-12
Add more descriptive error messages to CLI output


derekblank:
When encountering errors and displaying in the CLI, add descriptive human-readable error messages (in addition to displaying the error code).

PR #169, opened 2023-10-16
Add descriptive error messages to CLI output

derekblank:
Adds descriptive error messages to CLI command outputs (rather than returning only the error code). Also adds a confirm prompt to check if the branch already exists.

* https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/issues/163


derekblank:
@jhnstn This is a first-pass at how we might handle error messaging. It feels a bit more descriptive than just returning `pr, err`, and also helps a bit with self-documenting code. Let me know if you have any thoughts or suggestions for updating, or if there's any benefit to retaining all of the `return pr, err` statements. Otherwise, I can apply this error-message format to anywhere else we're checking for `err != nil`.

cli/pkg/release/gb.go
```diff
@@ -23,25 +23,32 @@ func CreateGbPR(version, dir string) (gh.PullRequest, error) {
...
 
-	console.Info("Checking if branch %s exists", branch)
 	exists, _ := gh.SearchBranch("gutenberg", branch)
 
 	if (exists != gh.Branch{}) {
-		console.Info("Branch %s already exists", branch)
-		return pr, nil
+		console.Warn("Branch %s already exists", branch)
+
+		cont := console.Confirm("Do you wish to continue?")
```
jhnstn, line 34: I like adding the confirm step here but we need more to get this working. 
On the other side of the `else` we are cloning the repo at `trunk` but if the branch exists we can clone off that. 

Also need to check if the PR already exists later on around line 133.

But all that can happen in a follow up PR
If you want we can return a "not implemented" error if someone confirms that working with the existing branch is ok. I'm not worried about that if we start on working out the existing branch work next

cli/pkg/release/gb.go
```diff
@@ -23,25 +23,32 @@ func CreateGbPR(version, dir string) (gh.PullRequest, error) {
...
+
 	} else {
 		console.Info("Cloning Gutenberg to %s", dir)
 
 		// Let's clone into the current directory so that the git client can find the .git directory
 		err := git.Clone(repo.GetRepoPath("gutenberg"), "--depth=1", ".")
+
 		if err != nil {
-			return pr, err
+			console.ExitError("There was an error cloning the Gutenberg repository.")
```
jhnstn: I don't think we should use `ExitError`  here or in the other changes. The common go practice is to return the error and let the caller decide how to handle it.  I know we "own" both this package and the command packages but still think we should bubble up errors as high as possible before exiting. In this case it would be in the command packages.

What we can do is decorate the error with more context. So in this case something like 
`return pr, fmt.Error("There was an error cloning the Gutenberg repository: %s", err)`

There are also some cases where the err does have some good debugging info. We keep that if all we do is extend the error message with more context.

jhnstn (changes requested):

derekblank:
Thanks for reviewing @jhnstn! I've updated to retain the `pr, err` return, but also include a more-specific error message along with the `err` output. Let me know what else we might want to include on this PR before merging. 👍

jhnstn (approved):
Looks good! Thanks for the updates.

PR #170, opened 2023-10-16
[CLI] improve exit handling

jhnstn:
Go will not call the deferred function when `os.Exit` is called with a non-zero value.

Currently the command is not cleaning up the temp directories if the run ends with an error.

These changes also move the error handling functions out of the console package. Ideally the console package would not have any side effects which this PR helps achieve. For now, our only entry points to the packages is from the /cmd directory, given that it made sense to move the exit handling functions there.

This PR can be tested as is by running the `release prepare gbm v1.2.3` command while that command is still not implemented. It will create a temporary directory but end in an error. 
It is expected that the temp dir is deleted.

### DIFF
--- a/cli/cmd/release/prepare/gb.go
+++ b/cli/cmd/release/prepare/gb.go
@@ -1,37 +1,40 @@
 package prepare
 
 import (
+	"errors"
+
 	"github.com/spf13/cobra"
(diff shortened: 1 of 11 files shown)


derekblank (approved):
LGTM - I was able to test successfully. Merging to incorporate in https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/169.

PR #175, opened 2023-10-17
Pass pr by value to preview function

jhnstn:
This should have landed in https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/158

Looks like trunk is broken without this change

### DIFF
--- a/cli/pkg/gh/gh.go
+++ b/cli/pkg/gh/gh.go
@@ -181,7 +181,7 @@ func labelRequest(rpo string, prNum int, labels []string) ([]Label, error) {
 	return resp, nil
 }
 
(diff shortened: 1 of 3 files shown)


cli/pkg/gh/gh.go
```diff
@@ -181,7 +181,7 @@ func labelRequest(rpo string, prNum int, labels []string) ([]Label, error) {
 	return resp, nil
 }
 
-func PreviewPr(rpo, dir string, pr *PullRequest) {
```
jhnstn, line 184: Since all the preview does is print the PR info, there is no need to pass the pointer.

derekblank (approved):
LGTM, merging to trunk. 🚀

PR #176, opened 2023-10-17
CLI: Make shell command interfaces public

jhnstn:
This allows passing the shell commands to helper functions.

Also rename factory functions with a `New` prefix to add clarity.


derekblank (commented):
I'm wondering if we could avoid the `New` prefix here -- from the perspective of others who haven't used the tool yet, it may not be providing enough context to know what was "old" in the first place. (This may also apply to the `deprecated` commands.) Could these commands just retain the current names without the `New` prefix?

derekblank (approved):
Oh, I understand now -- the `New` prefix is to indicate that these are factory functions, as the PR states. I thought it referred to them replacing a previous function. LGTM 🚀

PR #177, opened 2023-10-17
Create go.yml to run go workflows

jhnstn:
For now this just runs the tests but we can have it create builds if we want.


derekblank (approved):
LGTM. 🚀

PR #180, opened 2023-10-18
CLI: Refactor the temp directories in the commands

jhnstn:
This introduces a `workspace` package in the cmd directory.

Basic Usage:
`workspace := wp.NewWorkspace()`

`workspace.Dir()` returns the directory of the workspace. 
`workspace.Cleanup()` cleans up a temp directory if one was created
`workspace.Keep()` once called the clean up flows will not delete the temp directory.

The workspace can be disabled by setting the env `GBM_NO_WORKSPACE`. In this case a temp directory is not created. The rest of the script will use the current directory to fetch and update files.

Testing:

- The existing release commands should use and clean up temp directories as expected.


derekblank (approved):
LGTM. 🚀

PR #183, opened 2023-10-19
Add Release Notes and Related Changes when rendering Gutenberg Mobile PR template

derekblank:
Adds Release Notes and Related Changes when rendering Gutenberg Mobile PR template, including associated utils.

Based on previous examples of [getChangeLog](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/gbm-cli/cli/pkg/release/gbm.go#L159) and [getReleaseNotes](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/gbm-cli/cli/pkg/release/gbm.go#L199) to work with the [updated GBM PR template](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/gbm-cli/cli/templates/release/gbmPrBody.md) and ported to new utils/structure when necessary. 

#### To test:
From the `cli` directory, run the following against project repos forked to a workspace, e.g.:

```
GBM_WPMOBILE_ORG=derekblank GBM_WORDPRESS_ORG=derekblank go run main.go release prepare gbm 1.12.0
```

#### Todos to discuss:

- [x] **Passing `dir` to the get functions**: https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/183/files#r1366518576
- [x] **Getting `gbPr.Head.Sha` in the render function**: https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/183/files#r1366519773

### DIFF
--- a/cli/pkg/release/gbm.go
+++ b/cli/pkg/release/gbm.go
@@ -2,6 +2,9 @@ package release
 
 import (
 	"fmt"
+	"io"
+	"net/http"
+	"os"
 	"path/filepath"
 
 	"github.com/wordpress-mobile/gbm-cli/pkg/console"
@@ -212,6 +215,8 @@ func renderGbmPrBody(version string, pr *gh.PullRequest) error {
 			Version: version,
 		},
 	}
+	cl := getChangeLog(dir, pr)
+	rn := getReleaseNotes(dir, pr)
 
 	body, err := render.Render(t)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


cli/pkg/release/gbm.go
```diff
@@ -2,6 +2,9 @@ package release
 
 import (
 	"fmt"
+	"io"
+	"net/http"
+	"os"
```
derekblank, line 7: These imports are mostly scaffolding at this point just to keep the linter from throwing errors while we continue porting.

cli/pkg/release/gbm.go
```diff
@@ -49,11 +52,11 @@ func CreateGbmPR(version, dir string) (gh.PullRequest, error) {
...
-		console.Info("Set upstream to trunk", org)
-		err = git.SetUpstreamTo("trunk")
-		if err != nil {
-			return pr, err
-		}
+		// console.Info("Set upstream to trunk", org)
+		// err = git.SetUpstreamTo("trunk")
+		// if err != nil {
+		// 	return pr, err
+		// }
```
derekblank: I don't believe that it's actually necessary to set the upstream to `trunk` here -- likely it could be removed.

cli/pkg/release/gbm.go
```diff
@@ -203,20 +206,124 @@ func CreateGbmPR(version, dir string) (gh.PullRequest, error) {
 }
 
 func renderGbmPrBody(version string, pr *gh.PullRequest) error {
+	// TODO - replace "" with dir variable
+	cl := getChangeLog("", pr)
+	rn := getReleaseNotes("", pr)
```
derekblank, line 196: In the [first version of this code](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/gbm-cli/cli/pkg/release/gbm.go#L114-L115), `dir` was being passed to the `renderGbmBody` function. Adding a TODO to update it.

cli/pkg/release/gbm.go
```diff
@@ -203,20 +206,124 @@ func CreateGbmPR(version, dir string) (gh.PullRequest, error) {
...
+	var buff io.ReadCloser
+	cl := []byte{}
+
+	if dir == "" {
+		console.Warn("not implemented")
+
+		// TODO: find the best way to get the gbPr
+
+		// org, _ := repo.GetOrg("gutenberg")
+		// endpoint := fmt.Sprintf("https://raw.githubusercontent.com/%s/gutenberg/%s/packages/react-native-editor/CHANGELOG.md", org, gbPr.Head.Sha)
```
derekblank, line 253: We should determine the cleanest way to get `gbPr.Head.Sha` here, or refactor this if it's not needed.

jhnstn: The `gh` search should work here. Something like what we are doing to find the GBM pr for the integration PRs
here: https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/trunk/cli/pkg/release/integrate.go#L144-L146

cli/pkg/release/gbm.go
jhnstn: Good call. I wonder if we also need the `AddRemote` call above ?
I only remember running into git setup issues with the Gutenberg submodule. I would think since we are cloning from the remote repo, git would set that up for us by default.

cli/pkg/release/gbm.go
jhnstn: That search might not actually work _as-is_ . Searching GH for PRs does not return all of the PR info. We have to make another request once we know the PR number.
I made some improvements to PR searching [here](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/185/files#diff-7c6debd75d355092d930ab11518508e315ce45190d4bd377124d828eaa70d908)

jhnstn (approved):
Nice work!
This looks good to merge as-is. 
I think we can address the TODOs in follow up prs

PR #185, opened 2023-10-21
CLI: update integrate

jhnstn:
This updates the integrate command to handle cases when the release branches or PR exist. In these cases it will update the existing work.

It also includes a flow to check if the Gutenberg build is ready for each platform by checking the GBM pr status checks.
Note: since testing this flow is very difficult on a fork setup, the check is behind a "production" flag.

This PR also introduces a `status` command to check the status of a release `gbm release status {version}`

Testing:
Run one of the integrations on a clean fork (i.e. one with out a release PR or related branches)
Close the PR but keep the branches
Run the command again and note that the PR is recreated

Note: IOS tests must be made against a recent release ( `v1.106.0` for example ) otherwise `rake dependencies` will fail.

TODO : This will not recreate the "after" branches if the pr exists.


cli/pkg/release/integrate/integrate.go
```diff
@@ -0,0 +1,238 @@
...
+
+	"github.com/wordpress-mobile/gbm-cli/pkg/console"
+	"github.com/wordpress-mobile/gbm-cli/pkg/gbm"
+	"github.com/wordpress-mobile/gbm-cli/pkg/gh"
+	"github.com/wordpress-mobile/gbm-cli/pkg/render"
+	"github.com/wordpress-mobile/gbm-cli/pkg/repo"
+	"github.com/wordpress-mobile/gbm-cli/pkg/shell"
+)
+
+type Integration interface {
```
jhnstn, line 15: I updated the release integration handler to use interfaces to handle the different platforms.

cli/pkg/shell/git.go
```diff
@@ -29,6 +31,11 @@ func (c *client) Switch(args ...string) error {
 }
 
 func (c *client) CommitAll(format string, args ...interface{}) error {
+
+	if c.IsPorcelain() {
```
jhnstn, line 35: For our purposes we don't need the commit to fail if there are no changes. Instead it just warns and continues on.

derekblank (approved):
LGTM. The `status` command is neat. 🚀  Went ahead and resolved the merge conflicts that were caused by merging https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/183. Leaving the PR open if you want to double-check. 👍

PR #188, opened 2023-10-23
Cli: Add 'prepare all' command 

### LINKED ISSUE #187: Add command to run gb and gbm prepare
This command will be used to create both the GB and GBM release prs. 

It should check to make sure it's not being run on a CI since it won't have access to the GB repo.

jhnstn:
Fixes #187 
Adds a `release prepare all` command which runs both gb and gbm prepare commands.
This can be used to replicate most of the current release script.

This change set also includes a few other minor updates:
- Add more info to the status command
- Remove redundant error return from `repo.getOrg` calls. (most of the errors are assigned to `_`)
- Fix a couple of nil pointer issues in the `release/gbm` package.

### DIFF
--- a/cli/cmd/release/integrate.go
+++ b/cli/cmd/release/integrate.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/wordpress-mobile/gbm-cli/cmd/utils"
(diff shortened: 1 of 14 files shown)


derekblank (approved):
LGTM. 🚀 I didn't encounter any errors when running `go run main.go release prepare all v1.106.0`.

PR #200, opened 2023-10-26
CLI: Use npm install when setting up Gutenberg

jhnstn:
Gutenberg has a package-lock linter to make sure the local package versions match what's in the package lock.
Running `npm ci` won't update the lock file but `npm i` will

This changes the npm command to use `install`

Testing:
- Try preparing Gutenberg for a release
- Verify that the root package-lock.json file is updated.


derekblank (approved):
LGTM. 🚀

PR #202, opened 2023-10-26
Cli/console cleanup

jhnstn:
Console cleanup
- Remove most of the newlines in the general log messages
- Clean up the PR preview so it's more readable


derekblank (approved):
Added `Log` as it was missing from the GitCmds interface.

Issue #153, opened 2023-10-06
Update GB prepare command to handle patch releases


jhnstn:
Blocked by https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/issues/151

Given a patch release the prepare command should:

 - Checkout GB at the prior patch version e.g. if cutting `v1.0.1` the script should checkout GB at the `v1.0.0` tag
 - Accept a set of commits to cherry pick into the patch release branch
 - Proceed with the release preparation

jhnstn:
In theory we should be able to update the change notes programmatically. I would expect that any patch fix would include a change note in the unreleased section. From there cherry picking should filter out any other unreleased change notes... in theory. 

Merging back to trunk will be an issue but that is true regardless of if we automate the change note updates.

jhnstn:
I'm not sure how we want to handle accepting the commits to cherry pick

First, do we want to use PR numbers ? It would be nice to have those for adding to release PRs but we should be able to derive them from the github api.

Also should we have a prompt like the current script? how often do we need to cherry pick multiple commits/PRs ?
We could have the command accept a list of PR numbers (or commits). I think that would be easier if we commonly only cherry pick one or two commits. 

However we'd want to have a confirmation prompt to maker sure the wrangler has requested the correct changes to cherry pick. In this case we should allow for the wrangler to edit the requested changes, at least add more if something was forgotten

I suppose for a first pass we would add the confirm step and if it's not right the wrangler and exit the command and run it again with the correct PR numbers (or commits)

PR #204, opened 2023-10-26
CLI: Update version management

jhnstn:
This adds a `semver` package to handle the version management.
This is a refactor to lay the foundation for #153 where we will need to validate patch versions

Note: I looked into the go [semver](https://pkg.go.dev/golang.org/x/mod/semver) package but it surprisingly does not handle patches very well. But in the "go" way I just wrote some functions to handle versions the way we need to. 

Since most of the `pkg/utils` package was mostly for versioning, I also moved the rest of that package into new or exiting packages:
- `utils/utils.go#{version handling functions}` -> `pkg/semver` : cleaner interface for dealing with versions
- `utils/utils.go#NextReleaseDate` -> `cmd/render/utils` : It's only used in the render checklist command
- `utils/yq*` -> `pkg/yq` : Cleaner as its own package 
- `pkg/utils/change_logs.go` -> `pkg/release/utils.go` :  These are all release related


derekblank (approved):
LGTM 🚀

PR #205, opened 2023-10-30
Update CLI docs

derekblank:
Updates CLI docs for further information on available release process commands


cli/cmd/release/prepare/README.md
```diff
@@ -0,0 +1,29 @@
+# prepare
+
+The `prepare` command is responsible for generating the checklists that make up the release process.
```
jhnstn: This shouldn't say "checklists" right? Maybe it should say "pull requests" ?

jhnstn (approved):
Just need to change that one bit in prepare mentioning "checklists" . Otherwise looks good, thanks for keeping up on the docs!

I'll go ahead and approve it now

cli/cmd/release/prepare/README.md
derekblank: Good catch, updated 😎

Issue #194, opened 2023-10-25
Update Release Checklist with new CLI flow


derekblank:
Once the CLI tool is used for handling releases, we'll need to update the Release Checklist template with the new release steps (i.e., many can be deleted due to automation, some need to be added, etc).

PR #206, opened 2023-10-31
Update Release Checklist template for CLI

derekblank:
Updates the Release Checklist template for the new (and unused) steps to follow the new CLI tool process.

We can continue updating the template as we automate steps.

* https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/issues/194

### DIFF
--- a/cli/Releasing.md
+++ b/cli/Releasing.md
@@ -87,7 +87,6 @@ At the same time there could also be a regular release going on for example for 
 
 ### Automation script differences
 
-1. Before running the script switch to the relevant branch to cut from in gutenberg-mobile repo.
 1. Run the CLI tool as normal
 1. When asked by the script enter the relevant branch names to cut from (to target) in other repos.
 1. If a commit that is fixing the issue is already merged to gutenberg, when asked by the script enter the commit hash to be cherry-picked.
--- a/cli/Releasing.md
+++ b/cli/Releasing.md
@@ -1,0 +1,126 @@
+# Release Checklist Template
+
+The Release Checklist Template can be generated with the following command: (replacing `X.XX.X` with the applicable release number):
+
+```
+go run main.go render checklist -v X.XX.X --message "This hotfix addresses ISSUE_OR_CRASH in ORIGINAL_VERSION. See P2_PR_SLACK_ETC_LINK_OR_LINKS for the rationale behind the decision to make it."
+```
+
+The `--c` flag will copy the output to your clipboard. To view the output in the command line, simply remove the `--c` flag.
+
+
+If the release is a betafix/hotfix, include a message explaining briefly the reason for releasing the fix.
+**Example:** `This hotfix addresses ISSUE_OR_CRASH in ORIGINAL_VERSION. See P2_PR_SLACK_ETC_LINK_OR_LINKS for the rationale behind the decision to make it.`
+
+
+# Different types of releases
+
+## Best practices
+
+It's best practice to use the automation script (mentioned in the release template above) for all releases types (alpha, regular, betafix, hotfix). When wrangling a betafix or hotfix, it's important to merge the fix to Gutenberg `trunk` independently of the release process. When the release is cut (by the automation script) the commit(s) that make up the betafix or hotfix should then be cherry-picked onto the Gutenberg release branch.
+
+## 1. Alpha
+
+### When
+
+Whenever a build is needed for testing (usually a few days prior to a Regular release)
+
+### Branches
+
+For example, when the next release will be `1.11.0`.
+
+| Repo             | Cut From | Branch Name                               |
+| ---------------- | -------- | ----------------------------------------- |
+| gutenberg        | trunk    | rnmobile/release_1.11.0-alpha1            |
+| gutenberg-mobile | trunk    | release/1.11.0-alpha1                     |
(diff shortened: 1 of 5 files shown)


cli/Releasing.md
```diff
@@ -0,0 +1,129 @@
...
+| gutenberg        | trunk    | rnmobile/release_1.11.0-alpha1            |
+| gutenberg-mobile | trunk    | release/1.11.0-alpha1                     |
+| WPAndroid        | trunk    | gutenberg/integrate_release_1.11.0-alpha1 |
+| WPiOS            | trunk    | gutenberg/integrate_release_1.11.0-alpha1 |
+
+### Automation script differences
+
+Compared to a Regular release, the differences here are:
+
+- When the script asks for the new version number, don't forget to add the `-alpha` suffix (e.g. `1.11.0-alpha1`).
```
derekblank, line 46: @jhnstn Do we still do alpha releases? I'm not sure we need to keep this section.

jhnstn: We still do. There are cases where we introduce dependent code via the native bridge in Gutenberg into the host apps. When that happens we create an alpha release so the editor does not break in the main apps. Doesn't happen often however.

I do think we need to formalize when an alpha release is needed. There might be other reasons besides the one I just mentioned.

jhnstn: Also not sure how the CLI will handle the alpha version right now. I don't think it will work as-is.

cli/Releasing.md
```diff
@@ -0,0 +1,129 @@
...
+| Repo             | Cut From                | Branch Name                        | Merging To                                                       |
+| ---------------- | ----------------------- | ---------------------------------- | ---------------------------------------------------------------- |
+| gutenberg        | rnmobile/release_1.11.0 (tag) | rnmobile/release_1.11.1            | trunk & (maybe also) rnmobile/release_1.12.0                     |
+| gutenberg-mobile | v1.11.0 (tag)          | release/1.11.1                     | trunk & (maybe also) release/1.12.0                              |
+| WPAndroid        | release/22.2.0          | gutenberg/integrate_release_1.11.1 | release/22.2.0 & (maybe also) gutenberg/integrate_release_1.12.0 |
+| WPiOS            | release/22.2.0          | gutenberg/integrate_release_1.11.1 | release/22.2.0 & (maybe also) gutenberg/integrate_release_1.12.0 |
+
+### Automation script differences
+
+1. Before running the script switch to the relevant branch to cut from in gutenberg-mobile repo.
```
jhnstn: This step is not relevant to the CLI since we do fresh clones for each release.

cli/Releasing.md
```diff
@@ -0,0 +1,129 @@
...
+1. Run the CLI tool as normal
+1. When asked by the script enter the relevant branch names to cut from (to target) in other repos.
+1. If a commit that is fixing the issue is already merged to gutenberg, when asked by the script enter the commit hash to be cherry-picked.
```
jhnstn, line 92: This is still WIP but the CLI will accept a list of PRs via a pr flag ( `--prs` ). I Was considering an interactive prompt but we usually only include a single pr in a patch release I figure we don't need the prompt.

cli/templates/checklist/checklist.html
```diff
@@ -37,13 +37,12 @@ <h3>Before the Release (Tuesday)</h3>
 <h3>Create the Release{{ if .Scheduled }} (Thursday) {{end}}</h3>
 <!-- /wp:heading -->
 
-{{ Task `Verify that <code>gutenberg-mobile/RNTAztecView.podspec</code> and <code>gutenberg-mobile/gutenberg/packages/react-native-aztec/RNTAztecView.podspec</code>
- refer to the same <code>WordPress-Aztec-iOS</code> version and are pointing to a stable, tagged release (e.g. 1.14.1). If they are not, we may need to <a href="#create-a-new-aztec-release">create a new Aztec</a> release.` }}
+{{ Task `Clone the <a href="https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile">CLI</a> or pull the latest version if you have already cloned it.` }}
```
jhnstn: We shouldn't use clones of the CLI. Instead we should suggest installing the [latest release](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/releases)

cli/templates/checklist/checklist.html
```diff
@@ -37,13 +37,12 @@ <h3>Before the Release (Tuesday)</h3>
...
 
-{{ Task `Clone the <a href="https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile">release scripts</a> or pull the latest version if you have already cloned it.` }}
+{{ Task `Review the <a href="https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/develop/cli/Releasing.md">release script instructions</a>.
```
jhnstn: I'm not sure how relevant this is given the improvements we've made with the CLI. Perhaps we need to add this for patch releases but I don't think we need this for scheduled releases.

derekblank:
@jhnstn Updated from most of the comments - let me know if we think there are further areas to update to make the Release Checklist accurate. 👍

jhnstn (approved):
Looks good!

PR #207, opened 2023-11-01
Use FindGbReleasePr to update GBM release notes from CHANGELOG

derekblank:
Replaces a previous TODO item to return the CHANGELOG from the Gutenberg PR.

Leverages `FindGbReleasePr` to pass the PR version number and return the CHANGELOG to the `prepare` command as `cl`.

### DIFF
--- a/cli/pkg/release/gbm.go
+++ b/cli/pkg/release/gbm.go
@@ -221,20 +221,17 @@ func getChangeLog(dir string, gbmPr *gh.PullRequest) ([]byte, error) {
 	cl := []byte{}
 
 	if dir == "" {
-		console.Warn("not implemented")
-		return cl, nil
+		gbPr, _ := FindGbReleasePr(gbmPr.ReleaseVersion)
 
-		// TODO: find the best way to get the gbPr
+		org := repo.GetOrg("gutenberg")
+		endpoint := fmt.Sprintf("https://raw.githubusercontent.com/%s/gutenberg/%s/packages/react-native-editor/CHANGELOG.md", org, gbPr.Head.Sha)
 
-		// org, _ := repo.GetOrg("gutenberg")
-		// endpoint := fmt.Sprintf("https://raw.githubusercontent.com/%s/gutenberg/%s/packages/react-native-editor/CHANGELOG.md", org, gbPr.Head.Sha)
-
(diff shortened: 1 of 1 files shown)


derekblank:
@jhnstn I still need to do some testing, but this should get us pretty close to returning the CHANGELOG from the Gutenberg PR. Let me know if you have any thoughts or suggestions on the approach.

cli/pkg/release/gbm.go
```diff
@@ -221,20 +221,17 @@ func getChangeLog(dir string, gbmPr *gh.PullRequest) ([]byte, error) {
 	cl := []byte{}
 
 	if dir == "" {
```
derekblank, line 223: This `if dir == ""` logic block came from the original `gbm-cli` branch ([reference](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/gbm-cli/cli/pkg/release/gbm.go#L164)), but I'm wondering if it's still needed. Currently, this logic would always choose this block as we're already passing dir as `""` directly. Addressing this would tackle this todo item and issue for handling the dir:

https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/0073331ed6fab1391f8b14c010737b034487d949/cli/pkg/release/gbm.go#L163-L164

* https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/issues/208

jhnstn: I'm not sure why I added the dir check in the original `gbm-cli` branch. I know it was trigger the fetch to the remote `githubusercontent` site but I'm not sure why we would need to do that if we have a cloned local version. 

It might had been for a status command or a preflight check but I'm not entirely sure we now need what's in the `if dir == ""` block. Instead we should be calling `getChangeLog` with the directory that holds the GBM code.

derekblank: Updated to call `getChangeLog` and `getReleaseNotes` with the `dir` value. I kept the original logic in case it's useful in the future.

jhnstn: > I kept the original logic in case it's useful in the future.

Perfect

jhnstn (approved):

Issue #209, opened 2023-11-02
Handle logic for branches that already exist


derekblank:
When running `prepare`, handle the prompt logic to continue or exit if a branch already exists.

PR #210, opened 2023-11-02
Add prompt logic to handle if branch already exists

derekblank:
Resolves:
* https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/issues/209

When running the `go run main.go release prepare all x.xx.x` command and the branch already exists on the remote repo, the user is prompted if they want to continue. This PR adds the logic to delete the remote branch and continue cloning:

#### PR Changes
- Adds `DeleteBranch` GitHub interface
- When running `prepare`, moves `git.Clone` commands outside of the `else` statement to continue the automation (and for DRY-ness)

#### To test
1. Create a release, e.g. `go run main.go release prepare all 1.108.0`
2. Once the PR is created and the branch exists on the remote repo, run the same command again
3. At the prompt, choose either  yes _(continue)_ or no _(exit)_
4. If yes _(continue)_, the remote branch should be deleted and the script resumes cloning the repo
5. If no _(exit)_, the script should stop running and the remote branch remains

### DIFF
--- a/cli/pkg/release/gb.go
+++ b/cli/pkg/release/gb.go
@@ -27,28 +27,32 @@ func CreateGbPR(version, dir string, noTag bool) (gh.PullRequest, error) {
 	if (exists != gh.Branch{}) {
 		console.Warn("Branch %s already exists", branch)
 
-		cont := console.Confirm("Do you wish to continue?")
+		cont := console.Confirm("Do you wish to continue? (The remote branch will be deleted.)")
 
 		if !cont {
 			console.Info("Bye 👋")
 			return pr, fmt.Errorf("exiting before creating PR")
 		}
-		return pr, fmt.Errorf("existing branch not implemented yet")
-	} else {
-		console.Info("Cloning Gutenberg to %s", dir)
 
-		// Let's clone into the current directory so that the git client can find the .git directory
-		err := git.Clone(repo.GetRepoPath("gutenberg"), "--depth=1", ".")
-
-		if err != nil {
-			return pr, fmt.Errorf("error cloning the Gutenberg repository: %v", err)
+		// Delete the branch on the GitHub repo
+		if err := gh.DeleteBranch("gutenberg", branch); err != nil {
+			return pr, fmt.Errorf("error deleting the branch: %v", err)
 		}
+	}
 
-		console.Info("Checking out branch %s", branch)
-		err = git.Switch("-c", branch)
-		if err != nil {
-			return pr, fmt.Errorf("error checking out the branch: %v", err)
-		}
+	console.Info("Cloning Gutenberg to %s", dir)
+
+	// Let's clone into the current directory so that the git client can find the .git directory
+	err := git.Clone(repo.GetRepoPath("gutenberg"), "--depth=1", ".")
+
+	if err != nil {
+		return pr, fmt.Errorf("error cloning the Gutenberg repository: %v", err)
+	}
+
+	console.Info("Checking out branch %s", branch)
+	err = git.Switch("-c", branch)
+	if err != nil {
+		return pr, fmt.Errorf("error checking out the branch: %v", err)
 	}
 
 	console.Info("Updating package versions")
--- a/cli/pkg/release/gbm.go
+++ b/cli/pkg/release/gbm.go
@@ -37,19 +37,30 @@ func CreateGbmPR(version, dir string) (gh.PullRequest, error) {
 
 	if (exists != gh.Branch{}) {
 		console.Info("Branch %s already exists", branch)
-		return pr, nil
-	} else {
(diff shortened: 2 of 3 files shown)


cli/pkg/release/gb.go
```diff
@@ -27,28 +27,32 @@ func CreateGbPR(version, dir string, noTag bool) (gh.PullRequest, error) {
 	if (exists != gh.Branch{}) {
 		console.Warn("Branch %s already exists", branch)
 
-		cont := console.Confirm("Do you wish to continue?")
+		cont := console.Confirm("Do you wish to continue? (The remote branch will be deleted.)")
```
jhnstn, line 30: I'm not sure we want to delete the branch. What happens to the existing pull request?
Does it close or does it pick up the new branch? 
I think the bigger issue is the tag. I don't think deleting the branch will delete the tag ( unless the github api is doing some non-core git stuff)

I was thinking we would switch to the branch and update the PR with any new changes.
I'm not sure what that would look like for the scheduled release on Gutenberg. Perhaps we merge trunk into the branch but I doubt we'll need that since we narrowed the window between "prepare" and "cut" ( we use to prepare on Tuesday and cut on Thursday)

For patches we might want the flexibility to add in additional PRs before we cut the release.

cli/pkg/release/gbm.go
```diff
@@ -37,21 +37,32 @@ func CreateGbmPR(version, dir string) (gh.PullRequest, error) {
...
 	if (exists != gh.Branch{}) {
 		console.Info("Branch %s already exists", branch)
-		return pr, nil
-	} else {
-		console.Info("Cloning Gutenberg Mobile to %s", dir)
-		err := git.Clone(repo.GetRepoPath("gutenberg-mobile"), "--depth=1", "--recursive", ".")
-		if err != nil {
-			return pr, fmt.Errorf("error cloning the Gutenberg Mobile repository: %v", err)
+		
+		cont := console.Confirm("Do you wish to continue? (The remote branch will be deleted.)")
```
jhnstn, line 41: Since we don't push tags during the prepare step here I think deleting is ok. I'm just curious about what happens to the existing PR. We might want to add something about what happens to the PR in the confirm message.

Ideally we only have one PR per release, closed or open. 

Still thinking it would be cleaner to switch to the existing branch and then see what happens.  The `git.CommitAll` function checks to see if there is anything to commit before trying so that should work as expected. We should probably check to see if the PR already exists but we might want to try updating the PR description since a second prepare run might give us more info. For instance we might have the integration PRs at that time.

jhnstn (commented):
Deleting the existing branch might be ok but I’m not sure what happens to an existing PR and I think the tagging on GB will get weird

cli/pkg/release/gb.go
derekblank: > Does it close or does it pick up the new branch?

GitHub automatically closes the PR when a branch is deleted, but not the tag. I had deleting the branch in my mind because whenever I encounter a release branch that already exists, it's been because the release script failed and I need to start the whole process over. 😅 I'll try switching to the branch and see what happens.

derekblank: When switching to the existing branch, the branch already has changes, so we get:

```
Ready to create the PR on derekblank/gutenberg? [y/n]: 
To github.com:derekblank/gutenberg
 ! [rejected]        HEAD -> rnmobile/release_1.109.0 (fetch first)
error: failed to push some refs to 'github.com:derekblank/gutenberg'
hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
[ERROR] error pushing the PR: exit status 1
```

So we'll need to think through the flow a bit further. We can either:

1. Switch to the existing branch, reset the existing changes, and run the current flow
2. Delete both the branch and the tag on the repo and start over
3. Clone the repo with the existing branch already checked out while cloning, e.g.: `git clone --branch release/x.xxx.x` 

I'm not sure which is the cleanest approach, or the most likely scenario we'd run into. WDYT? I haven't done a lot of patch releases, mostly scheduled releases, so I might not be aware of some of the more nuanced use cases. Can you think of other scenarios where the branch might already exist that we'd to account for?

jhnstn: hmm I haven't had a chance to try this out but did the switch command include the `-c` flag? 
`git switch {some_branch}` should fetch and checkout the HEAD of the remote branch. In that case there shouldn't be any rejected updates. The `-c` flag will create a new local branch which would cause issues when trying to push.

derekblank:
@jhnstn Closing this one, we can re-open it or create a new issue if we need to.

PR #213, opened 2023-11-03
CLI: Fix how nvm installs node

jhnstn:
This fixes how nvm is used to load the correct node version.

I believe previous attempts to fix this were masked by setting the global node to match the required version.

## Testing
- Check to make sure the default node version does not match the one in the Gutenberg .nvmrc file (fork or main repo works)
- Run the `prepare gb {version}` command
- Notice after the "Setting up Gutenberg node environment" that nvm logs that it found a .nvmrc file and switched node versions
- Confirm that the node set up succeeds

### DIFF
--- a/cli/pkg/utils/utils.go
+++ b/cli/pkg/utils/utils.go
@@ -6,6 +6,8 @@ import (
 	"os/exec"
 	"regexp"
 	"time"
+
+	"github.com/wordpress-mobile/gbm-cli/pkg/console"
 )
 
 func ValidateVersion(version string) bool {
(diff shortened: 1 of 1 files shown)


cli/pkg/utils/utils.go
```diff
@@ -45,12 +47,16 @@ func SetupNode(dir string) error {
 
 	// Check for nvm
 	if os.Getenv("NVM_DIR") != "" {
-		cmd = exec.Command("bash", "-l", "-c", "$NVM_DIR/nvm.sh", "use")
+		cmd = exec.Command("bash", "-l", "-c", ". $NVM_DIR/nvm.sh && nvm use")
```
jhnstn, line 15: I realized that the first command `$NVM/nvm.sh` just loads nvm in the terminal. The following `use` call is also needed.

derekblank (approved):
LGTM! I was able to follow the testing instructions successfully. 🚀

PR #215, opened 2023-11-06
[CLI] Fix module path

jhnstn:
This updates the module path so that it correctly matches the current repo path. It also renames the directory to give more context to the binary
With this change the CLI tool can be installed via: 
`go install ithub.com/wordpress-mobile/release-toolkit-gutenberg-mobile/compare/gbm-cli@latest`

Which will install the latest release in the users `GOPATH`. The executable will then be available with `gbm-cli`


derekblank (approved):
Merged `trunk` to resolve the README conflicts created when https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/pull/206 was merged.  Feel free to revert the commit if it breaks something on your end. Otherwise, looks good to me. 🚀

PR #218, opened 2023-11-08
[CLI] Update docs and add cli versioning

jhnstn:
This updates the docs to reflect the new CLI tool. 
It also introduces a version check to the cli tool

## Highlights

- Consolidate the `Releasing.md` files
- Add more information on installing and using the cli tool
- Add a warning if using an outdated version of the cli
- Remove the bin directory in favor of adding binaries to the releases

## Testing

- Run `GBM_WPMOBILE_ORG=wordpress-mobile go run main.go --version` 
- Expect to see a warning about the version: "[WARN] You are running an older version of the CLI. Please update to v1.1.0"
- Verify that the version printed is "v1.2.0"

Note: The message might seem off since the published version is behind the one in this PR. This is expected since this PR first introduces the cmd versioning. Once this is merged, a new release will be created and the message will make more sense.


derekblank (approved):
LGTM! Thanks for updating the docs. 🚀

PR #219, opened 2023-11-08
[CLI] Don't create a temp workspace on CI

### LINKED ISSUE #174: CLI: Update how we handle cleaning up the temp directories
Cleaning up the temp dir after creating the gbm release pr can take awhile. It looks like the command is just hanging after running.
We should add a message to warn folks that deleting the temp dir can take awhile. 

We might also need to rethink how we handle the temp directories on the gbm and integration related commands. Mostly since we plan to run these on a CI. For starters we probably don't need to clean up the temp dirs since the VM is likely to go away after the CI run. But we might be able to use the home directory that the CI runs in.

jhnstn:
Fixes #174 

We won't need a temp directory on a CI assuming we are spinning up a new VM on each run (like what happens on Github Workflows)

Instead we can just use the directory set up by the CI to run the commands.

This adds a check for the `CI` env before creating a temp directory. Since a temp directory is not created, nothing is deleted after the command runs

## Testing 
- Create a build to some empty directory e.g. `go build ~/some-test-directory/gbm-cli`
- `cd` into that directory and create a new empty directory `gb`
- From the directory created in the last step run `CI=true ../gbm-cli release prepare gb {version} --no-tag`
- Notice a message about not creating a temp directory
- Exit before creating the PR
- Note that Gutenberg is cloned to the current directory 
- Verify that nothing was deleted at the end of the command run

### DIFF
--- a/gbm-cli/cmd/workspace/workspace.go
+++ b/gbm-cli/cmd/workspace/workspace.go
@@ -30,6 +30,13 @@ func NewWorkspace() (Workspace, error) {
 		console.Info("GBM_NO_WORKSPACE is set, not creating a workspace directory")
 		w.disabled = true
 	}
+
+	if _, ci := os.LookupEnv("CI"); ci {
+		console.Info("CI environment detected, not creating a workspace directory")
+		w.disabled = true
+		w.dir = "."
+	}
+
 	if err := w.create(); err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


derekblank (approved):
LGTM -  was able to follow the testing command sucessfully. 🚀

PR #220, opened 2023-11-08
[CLI] call `nvm use` before each npm call on CI

jhnstn:
While testing the prepare command in a Github workflow I noticed that any call to `exec.Command` seemed to run in a separate process.
The result is that `nvm` needs to be ran before each npm command. Without that the workflow falls back to the default system node.

## Testing 
The real test is to see if the changes work on a github workflow but we can test locally to make sure it's working.

Steps to check locally:
- Create a build to some empty directory e.g. go build `~/some-test-directory/gbm-cli`
- `cd` into that directory and create a new empty directory `gb`
- From the directory created in the last step run `CI=true ../gbm-cli release prepare gb {version} --no-tag`
- Confirm that command succeeds


derekblank (approved):
Good catch! LGTM. 🚀

PR #222, opened 2023-11-08
[CLI] Updates for CI workflows

jhnstn:
Switches the git clone command to use `https://` instead of `git@`. This will allow us to clone the repos without ssh keys. 

This is needed to run the commands on CI where we will use a Github Token to access the repos

Also includes a check to return true from `console.Confirm` and skips printing the PR Preview if running on CI 

## Testing
Confirm that any of the release command can still clone and push to the repos.

### DIFF
--- a/gbm-cli/pkg/repo/repo.go
+++ b/gbm-cli/pkg/repo/repo.go
@@ -65,3 +65,8 @@ func GetRepoPath(repo string) string {
 	org := GetOrg(repo)
 	return fmt.Sprintf("git@github.com:%s/%s", org, repo)
 }
+
+func GetRepoHttpsPath(repo string) string {
+	org := GetOrg(repo)
+	return fmt.Sprintf("https://github.com/%s/%s", org, repo)
+}
--- a/gbm-cli/pkg/repo/repo.go
+++ b/gbm-cli/pkg/repo/repo.go
@@ -53,7 +53,7 @@ func GetOrg(repo string) string {
 	case WordPressAndroidRepo:
 		fallthrough
 	case ReleaseToolkitGutenbergMobileRepo:
-		fallthrough
+		return "wordpress-mobile"
 	case WordPressIosRepo:
 		return WpMobileOrg
 	default:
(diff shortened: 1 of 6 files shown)


derekblank (changes requested):
When testing this, it looks like it's picking up my env vars to set the release-toolkit-gutenberg-mobile org URL:

```
release-toolkit-gutenberg-mobile/gbm-cli on  cli/clone-with-https [$] via 🐹 v1.21.2 
➜ yes | GBM_WPMOBILE_ORG=derekblank GBM_WORDPRESS_ORG=derekblank go run main.go release prepare gb 1.116.0

HTTP 404: Not Found (https://api.github.com/repos/derekblank/release-toolkit-gutenberg-mobile/releases/latest)

exit status 1
```

However, `case ReleaseToolkitGutenbergMobileRepo` is set to `fallthrough` so this seems odd. Should we force the `ReleaseToolkitGutenbergMobileRepo` org to always be `wordpress-mobile`?

jhnstn:
Oh hmm, yeah I suppose we should force it to `wordpress-mobile` I've been testing with a release in my fork toolkit so I haven't seen this. Thanks for catching it!

gbm-cli/pkg/repo/repo.go
```diff
@@ -53,7 +61,7 @@ func GetOrg(repo string) string {
 	case WordPressAndroidRepo:
 		fallthrough
 	case ReleaseToolkitGutenbergMobileRepo:
-		fallthrough
+		return ToolkitOrg
```
derekblank, line 64: Initially I hardcoded this string to "wordpress-mobile", but thought we could do one better and add `GBM_TOOLKIT_ORG` as an env so we can still point at a fork when needed. It defaults to `wordpress-mobile`.

derekblank (approved):
Added `GBM_TOOLKIT_ORG` as an env to bypass the org/fork error, and was able to test successfully. 👍 Feel free to modify `gbmToolkitOrg` if it can be improved (or scrap it if you want), but otherwise the functionality LGTM. 🚀

PR #224, opened 2023-11-09
[CLI] Add watch feature to release status

jhnstn:
Adds a `--watch` and `--time` flag to the status command.
Makes is easier to know when the platform builds are ready

I also cleaned up some of the not so helpful warning messages

## Testing
Try the new watch flag on an existing release


derekblank (approved):
Looks great! 🚀

PR #229, opened 2023-11-28
[CLI] Fix Github Token handling

jhnstn:
This uses the same function `go-gh` uses to find the users Github auth token when cloning repos. 
Previously it was only looking for `GITHUB_TOKEN`.
This change will allow users to just run `gh auth login` to use the tool or it will work by supplying a github token via `GH_TOKEN` or `GITHUB_TOKEN` if `gh` is not installed

##Testing

### With out `gh`
1. Uninstall `gh` ( `brew uninstall gh`)
2. Make sure `gh` has not added anything to `~/.gitconfig` look for something like
```
[credential "https://github.com"]
    helper = !/opt/homebrew/bin/gh auth git-credential
```
4. Make sure `~/.config/gh/hosts.yml` does not have an `oauth_token` 
3. Try creating a test release `go run main.go release prepare gb <version> --no-tag`
4. Note that the command fails
5. Try using a [github PAT](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens) by adding that as an env : 
```
GH_TOKEN=<your github token> go run main.go release prepare gb <version> --no-tag
// or
GITHUB_TOKEN=<your gitub token> go run main.go release prepare gb <version> --no-tag
```
6. Verify that the changes can be pushed to github and that the PR is created

### With `gh` installed
1. re-install `gh` 
2. Try without logging in by repeating steps 5 and 6 above
3. Login in via `gh auth login` and **make sure to not run** `gh auth setup-git` 
4. Verify that `gh` has not modified `~./gitconfig`
5.  Try creating a test release without the exported token:  `go run main.go release prepare gb <version> --no-tag`
6. Verify that the changes can be pushed and that the PR is created

### DIFF
--- a/gbm-cli/cmd/root.go
+++ b/gbm-cli/cmd/root.go
@@ -8,7 +8,7 @@ import (
 	"github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/gbm-cli/pkg/gh"
 )
 
-const Version = "v1.3.0"
+const Version = "v1.4.0"
 
 // rootCmd represents the base command when called without any subcommands
 var rootCmd = &cobra.Command{
--- a/gbm-cli/go.mod
+++ b/gbm-cli/go.mod
@@ -5,6 +5,7 @@ go 1.21
 toolchain go1.21.2
 
 require (
+	github.com/davecgh/go-spew v1.1.1
 	github.com/fatih/color v1.15.0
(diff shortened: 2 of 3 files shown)


derekblank (approved):
I was able to follow the testing steps successfully. 🚀

PR #241, opened 2024-01-26
CLI: Fix nvm setup

### LINKED ISSUE #240: [CLI] Fix logic to load nvm when calling npm commands
The script will set up `nvm` on every `npm` call when running on CI ([source](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/trunk/gbm-cli/pkg/shell/cmds.go#L37)).

The issue is that Go runs commands with `/bin/bash` by default so unless `nvm` is set up to run in `bash`, the `npm` calls that the script makes will not load `nvm` and fall back to the systems global node/npm.

Locally the fix is to make sure that `nvm` is set up to work with bash by adding the "[nvm snippet](https://github.com/nvm-sh/nvm#installing-and-updating)" to `~/.bash_profile`

jhnstn:
Fixes #240 

This sets up the npm command to correctly use nvm. 
It also provides a clear space to introduce other node managers.

## Testing 

Follow the [Testing](https://github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/blob/trunk/gbm-cli/Testing.md) guide to set up the tool to use forked repos.

Prerequisite: nvm installed 

### Testing with nvm
- Verify nvm is not loaded in `bash_profile` , e.g.
```
$ bash -l 
$ nvm  
bash: nvm: command not found
```
- Make sure your default node in nvm is not v20
- Try running `go run main.go release prepare gb {Version}` 
- Verify that the command reaches the "create PR" confirmation step

### Testing without nvm
- Make sure your default node in nvm is not v20
- Try `NVM_DIR= go run main.go release prepare gb {Version}`
- This should fail with unsupported engine error
- Set the default node to v20 e.g. `nvm alias default v20`
- Try `NVM_DIR= go run main.go release prepare gb {Version}` again 
- Verify that the command reaches the "create PR" confirmation step


dcalhoun (commented):
I tested this while using `asdf-vm` without `nvm` installed. The script works as expected, utilizing the "system" Node.js. 

Whenever my default Node.js is set to the desired version, the script uses it. I may explore how I might tweak my system bash configuration to expose my local, project-specific Node.js version controlled by `asdf-vm` to the script environment. 

I'll let others provide feedback on the `nvm` support, but this looks ready to merge from. my perspective.

dcalhoun (approved):
Optimistically approving `nvm` support even though I did not test it directly. Others can test and provide feedback in retrospect, if desired.

PR #247, opened 2024-03-08
[CLI] Add ability to self update binary

jhnstn:
This adds the ability to self update the cli tool.

### Why?
Apple requires code signing for all binaries, making it tricky to download the `gbm-cli` tool directly from the github releases pages. This change will make it easy for release wranglers to update the tool.

### Testing

- Change the version in `./gbm-cli/cmd/root.go` to a lower version (current is 1.5.0), example patch:

```patch
diff --git a/gbm-cli/cmd/root.go b/gbm-cli/cmd/root.go
index e921336..a7fefb9 100644
--- a/gbm-cli/cmd/root.go
+++ b/gbm-cli/cmd/root.go
@@ -8,7 +8,7 @@ import (
 	"github.com/wordpress-mobile/release-toolkit-gutenberg-mobile/gbm-cli/pkg/console"
 )
 
-const Version = "v1.5.0"
+const Version = "v1.4.0"
 
 // rootCmd represents the base command when called without any subcommands
 var rootCmd = &cobra.Command{
```
- Run `go build -o ./bin/gbm-cli` 
- Run `./bin/gbm-cli --version`
- Cancel the option to update and note the version number. It should be the value added in the first step
- Run `./bin/gbm-cli --version` again 
- Accept the offer to update
- Note that the version message shows the latest version in Github
- Run `./bin/gbm-cli --version` again 
- Note that there is no prompt to update

### DIFF
--- a/gbm-cli/cmd/release/status.go
+++ b/gbm-cli/cmd/release/status.go
@@ -2,6 +2,7 @@ package release
 
 import (
 	"fmt"
+	"reflect"
 	"time"
 
 	"github.com/fatih/color"
@@ -39,7 +40,7 @@ var StatusCmd = &cobra.Command{
 			// Check to see if the release has been published
 			rel, _ := release.GetGbmRelease(version)
 
-			if (rel != gh.Release{}) {
+			if (!reflect.DeepEqual(rel, gh.Release{})) {
 				console.Print(heading, "\n🎉 Release %s has been published!\n %s\n", version, rel.Url)
 			}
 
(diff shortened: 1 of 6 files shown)


gbm-cli/cmd/release/status.go
```diff
@@ -39,7 +40,7 @@ var StatusCmd = &cobra.Command{
 			// Check to see if the release has been published
 			rel, _ := release.GetGbmRelease(version)
 
-			if (rel != gh.Release{}) {
+			if (!reflect.DeepEqual(rel, gh.Release{})) {
```
jhnstn, line 43: This is the only possible regression since the simple equality won't work with the array of `ReleaseAsset` added to the `Release` struct. This can be verified by running a status check on a published release e.g.
```
go run main.go release status 1.114.1
```

PR #252, opened 2024-04-25
Remove login shell flag so that environment variables are used.

twstokes:
Environment variables from the interactive shell weren't being included when the "login shell" [flag](https://www.gnu.org/software/bash/manual/html_node/Invoking-Bash.html) was supplied to Bash. Utilities like `rbenv` depend on environment variables to override the system version of Ruby.


geriux (approved):
LGTM! 🚀

